PR #60, opened 2026-03-02
Add compute_eigensystems parameter to opt out of eigendecomposition

### LINKED ISSUE #51: Eigensystems always computed for rank-2 tensors; no way to opt out
## Issue

Any rank-2 tensor in \`wanted\` always triggers eigensystem computation:

\`\`\`python
for name in _RANK2:
    if name in wanted and all_raw[name]:
        eig_raw[name] = calculate_eigensystem(all_raw[name])
\`\`\`

There is no way to:
- Request the raw 3×3 tensor matrix without paying the eigendecomposition cost
- Request *only* eigensystems without also storing the raw tensor
- Disable eigensystems entirely for a batch job that only needs scalars and vectors

For pipelines processing thousands of labels per image, the six \`np.linalg.eigh\` calls per label can dominate runtime.

## Proposed options

**Option A (minimal):** Add a \`compute_eigensystems=True\` boolean parameter to \`minkowski_tensors\`.

**Option B (fine-grained):** Support explicit eigensystem keys in the \`compute\` list, e.g.:
\`\`\`python
compute=['w020', 'w020_eigvals']    # tensor + eigvals only
compute=['w020_eigvals']            # eigvals without storing the matrix
\`\`\`

**Option C (document only):** Document the current behaviour and defer the opt-out to a later milestone.

## Decision needed

Which option to adopt should be decided as part of M0 API stabilisation, even if implementation is deferred. Option A is lowest risk for M0.

yajushikhurana:
Resolves #51.

## Summary

- Adds `compute_eigensystems=True` to `minkowski_functionals()` and `minkowski_functionals_from_label_image()`. Default keeps existing behaviour unchanged.
- When `False`, rank-2 tensor matrices (3×3) are still computed and returned — only the eigendecomposition (`*_eigvals` / `*_eigvecs` keys) is skipped, avoiding six `np.linalg.eigh` calls per label.
- Raises `ValueError` if any beta-derived quantity (e.g. `'beta'`, `'w020_beta'`) is requested alongside `compute_eigensystems=False`, anticipating the hard dependency introduced by #1.

## Test plan

- [ ] `compute_eigensystems=False` omits `*_eigvals` / `*_eigvecs` from output
- [ ] `compute_eigensystems=False` still returns raw 3×3 tensor matrices
- [ ] `compute_eigensystems=False` works cleanly with scalar-only `compute` lists
- [ ] `compute_eigensystems=True` (default) preserves existing eigensystem output
- [ ] `beta` + `compute_eigensystems=False` raises `ValueError` with clear message
- [ ] `*_beta` suffix + `compute_eigensystems=False` raises `ValueError`
- [ ] `minkowski_functionals_from_label_image` threads the parameter through correctly
- [ ] All 35 existing tests continue to pass

### DIFF
--- a/pykarambola/api.py
+++ b/pykarambola/api.py
@@ -266,9 +266,15 @@ def minkowski_functionals_from_label_image(
         ``None``: use the origin.
         ``(3,)`` array: use an explicit point for all labels.
     compute : str or list of str
-        Passed through to :func:`minkowski_functionals`.
+        Which functionals to compute. ``'standard'`` returns the 14 base
+        functionals; ``'all'`` additionally computes ``w103``, ``w104``,
+        and spherical Minkowski summaries (``msm_ql``, ``msm_wl``); a list
+        of names selects specific quantities.
     compute_eigensystems : bool, optional
-        Passed through to :func:`minkowski_functionals`.
+        If False, eigenvalues and eigenvectors for rank-2 tensors are
+        skipped (``*_eigvals`` / ``*_eigvecs`` keys are omitted from each
+        label's result dict), avoiding six ``np.linalg.eigh`` calls per
+        label. Default is True.
 
     Returns
     -------
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -237,7 +237,23 @@ class TestComputeEigensystems:
             assert f'{name}_eigvals' in result
             assert f'{name}_eigvecs' in result
 
+    def test_single_rank2_tensor_omits_eigensystem(self):
+        # Requesting a single rank-2 tensor with compute_eigensystems=False
+        # should return the tensor matrix but suppress its eigvals/eigvecs.
+        result = minkowski_functionals(
+            self.verts, self.faces,
+            compute=['w102'],
+            compute_eigensystems=False,
+        )
+        assert 'w102' in result
+        assert result['w102'].shape == (3, 3)
+        assert 'w102_eigvals' not in result
+        assert 'w102_eigvecs' not in result
+
     def test_beta_with_false_raises_value_error(self):
+        # 'beta' is not yet implemented as a functional, but the guard fires
+        # proactively as a forward-compatibility check so that once beta lands
+        # it cannot be requested without eigensystems.
         with pytest.raises(ValueError, match="compute_eigensystems=True"):
             minkowski_functionals(
                 self.verts, self.faces,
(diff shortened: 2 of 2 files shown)


keisuke-ishihara (commented):
## Review

### Test coverage gaps

KI: from a quick glance, I understood that the new tests were testing combintions of parameters passed to `minkowski_functionals()`, but i wasn't sure if it covered all combinations. I asked Claude to look into this, and it found the following.

The new `TestComputeEigensystems` class covers the core cases well, but two combinations are missing:

| `compute` value | `compute_eigensystems=False` |
|---|---|
| `'standard'` | ✅ covered |
| scalar-only list | ✅ covered |
| `'beta'` / `'*_beta'` | ✅ raises `ValueError` |
| `'all'` | ❌ not tested |
| single rank-2 tensor (e.g. `['w102']`) | ❌ not tested |

The most important gap is **`compute='all'` + `compute_eigensystems=False`**: worth adding a test that verifies `w103`, `w104`, `msm_ql`/`msm_wl` are still returned while `*_eigvals`/`*_eigvecs` are suppressed.

KI: I agree that we might as well test all combinations. Can you add new test functions to cover the missing combos?

---

KI: The following points were raised by Claude, when I asked it to review. Point 1 is a good warning to keep in mind. Point 2, we could add a brief oneliner in the docstring. Point 3, a short comment might be worth adding.

### Other points

**1. Merge conflict with PR #61**
Both PRs touch `minkowski_functionals_from_label_image`. PR #61 changes the `center=` default from `'centroid'` to `'centroid_mesh'` and restructures the centroid block. Whichever merges second will need to rebase — worth coordinating the merge order.

**2. Terse docstring in `from_label_image`**
```python
compute_eigensystems : bool, optional
    Passed through to :func:`minkowski_functionals`.
```
This requires the reader to follow the cross-reference to understand the behaviour. A brief one-liner (e.g. *"If False, `*_eigvals`/`*_eigvecs` keys are omitted from the output."*) would make the docstring self-contained.

**3. Beta guard tests rely on `'beta'` being an unknown compute name**
`test_beta_with_false_raises_value_error` passes `compute=['w020', 'beta']`. Until issue #1 lands, `'beta'` is silently ignored by the `wanted` set — but the `ValueError` is correctly raised before computation starts, so the test passes for the right reason. A short comment in the test explaining this would avoid future confusion when `'beta'` becomes a valid compute name.

PR #82, opened 2026-03-10
Fix center parameter API and add object count (#73, #80)

### LINKED ISSUE #73: `center` parameter inconsistency between `minkowski_tensors` and `minkowski_tensors_from_label_image`
## Summary

The two public API functions handle centroid-based reference points in incompatible ways, and neither function gives users explicit control over whether the centroid is computed **per labeled object** or from the **entire mesh/image globally** when labels are present.

## Background: C++ karambola

pykarambola is a Python port of [karambola](https://github.com/morphometry/karambola). The C++ version supports:
- **Labeled mesh inputs** via `--labels` / `--no-labels` flags — Minkowski tensors are computed globally or per label within a single mesh
- **`--reference_centroid`** — computes tensors relative to the "Minkowski Vectors' center of gravity". The centroid is **not** a user-supplied coordinate; it is derived automatically from the mesh's own Minkowski functionals before tensor calculations run. In pykarambola this translates to each rank-2 tensor using its own per-tensor reference vector:

  | Tensor | Reference vector | Meaning |
  |--------|-----------------|---------|
  | `w020` | `w010 / w000`   | volume-weighted centroid (center of mass) |
  | `w120` | `w110 / w100`   | surface-area-weighted centroid |
  | `w220` | `w210 / w200`   | mean-curvature-weighted centroid |
  | `w320` | `w310 / w300`   | Gaussian-curvature-weighted centroid |

  That is, **4 distinct centroids per object**, one for each rank-2 tensor, rather than a single shared reference point.

- **Labels × reference_centroid interaction in C++:**
  - `--nolabels --reference_centroid`: centroid computed from the whole mesh, applied globally
  - `--labels --reference_centroid`: centroid computed **independently per label**, each label uses its own Minkowski-derived centroid

More broadly, which functionals require a reference point:

| Rank | Functionals | Reference point needed? |
|------|-------------|------------------------|
| Scalar (0) | `w000`, `w100`, `w200`, `w300` | No |
| Vector (1) | `w010`, `w110`, `w210`, `w310` | No — these *encode* position from the origin; their ratio to the scalar gives the centroid |
| Tensor (2) | `w020`, `w120`, `w220`, `w320` | Yes — vertex positions are shifted by `-ref_vec` before integration |
| Tensor (3) | `w103` | No — integrates normal vectors only; translation-invariant |
| Tensor (4) | `w104` | No — integrates normal vectors only; translation-invariant |

The C++ version does **not** have a geometric mesh centroid option. pykarambola should preserve C++ API compatibility while offering additional options where useful.

### `minkowski_tensors(verts, faces, labels=..., center=...)`

- `center='centroid'` corresponds to the C++ `--reference_centroid`: feeds already-computed prerequisite scalars/vectors into rank-2 tensor formulas, yielding 4 distinct per-tensor centroids. This matches C++ behavior.
- **However**, when `labels` is provided, `center` currently applies as a **single shared value for all labels** — there is no per-label centroid support, which breaks parity with C++ `--labels --reference_centroid`.
- `center='centroid_mesh'` does not exist in this function at all.
- No way to choose between per-label and global centroid when labels are present.

### `minkowski_tensors_from_label_image(..., center='centroid_mesh')`

- `center='centroid_mesh'` is a pykarambola addition with no C++ equivalent: computes the volume-weighted centroid via the divergence theorem (`w010 / w000`) — equivalently, the **center of mass** — and shifts vertices by `-centroid` before computation.
- Is naturally per-label because each label is meshed independently in this function.
- Has no equivalent `'reference_centroid'` option, making the two functions asymmetric in the opposite direction.

## Impact

1. A user with `(verts, faces, labels)` mesh data who wants **per-label geometric centroids** (equivalent to `centroid_mesh`) has no clean path — they must manually split the mesh by label and loop.
2. A user who wants to apply a **single global centroid across all labels** (e.g. center everything relative to the whole-image center of mass) has no built-in option in either function.
3. The name `center='centroid'` does not reflect its C++ origin as `--reference_centroid`, making the correspondence non-obvious.

### 1. Rename `center='centroid'` → `center='reference_centroid'`
In both functions, with a deprecation warning on the old name `'centroid'`. This aligns the option name directly with the C++ `--reference_centroid` flag.

### 2. Add `center='centroid_mesh'` to `minkowski_tensors`
`centroid_mesh` computes the volume-weighted centroid (`w010 / w000`) — the center of mass — and shifts all vertices by that value before computing tensors. This is the same concept as center of mass; the name `centroid_mesh` is used to keep both functions consistent with each other.

### 3. Make centroid scope explicit for the labeled case
When `labels` is provided, both `reference_centroid` and `centroid_mesh` should support two scopes via a new `center_scope` parameter:

| `center_scope` | Behavior |
|----------------|---------|
| `'per_label'` (default) | Centroid computed independently for each labeled sub-mesh; each label is shifted by its own centroid. Matches C++ `--labels --reference_centroid` behavior. |
| `'global'` | Centroid computed from the entire mesh (all labels combined), then applied as a single shift to all labels before dispatch. |

When no `labels` is provided, `center_scope` has no effect (single body, always global). Passing an explicit `center=(x, y, z)` array continues to work as a manual global shift and ignores `center_scope`.

### Summary of target API

| `center` value | `center_scope` | Behavior |
|----------------|----------------|---------|
| `None` | — | Origin (0,0,0) for all labels |
| `'reference_centroid'` | `'per_label'` | Per-label, per-tensor Minkowski centroid (C++ default with `--labels`) |
| `'reference_centroid'` | `'global'` | Single whole-mesh Minkowski centroid applied to all labels |
| `'centroid_mesh'` | `'per_label'` | Per-label geometric center of mass (volume-weighted) |
| `'centroid_mesh'` | `'global'` | Single whole-mesh center of mass applied to all labels |
| `(x, y, z)` array | — | Explicit single shift applied to all labels |

### LINKED ISSUE #80: Return object count for mesh and label-image inputs (labelled and unlabelled)
## Summary

Neither `minkowski_tensors()` nor `minkowski_tensors_from_label_image()` 
exposes the number of distinct objects in the input. For unlabelled inputs, 
connected components are never detected at all.

## Current behaviour

- **Mesh (`minkowski_tensors`):** `labels=None` silently merges all 
  disconnected components. When labels are provided, the caller must 
  inspect `len(result)` to infer the count.
- **Label image (`minkowski_tensors_from_label_image`):** a binary image 
  with multiple disconnected blobs under the same label value is treated 
  as a single object. Count must again be inferred from `len(result)`.

## Proposed fix

Add a `return_count=True` flag to both functions. When set, the return 
value becomes a `(results, n_objects)` tuple, where `n_objects` is the 
total number of connected components across all labels. The `results` 
dict is **unchanged** — no splitting or relabelling occurs.

Connected-component counting operates at the component level, not the 
label-value level: two disconnected blobs sharing the same integer label 
count as 2 objects, but their merged result dict entry is preserved as-is.

**Implementation differs by input type:**
- **Label image:** `skimage.measure.label` on each per-label binary mask.
- **Mesh:** face/vertex adjacency graph traversal on the per-label 
  face subgraph (or the whole mesh when unlabelled).

## Edge cases

- Empty mesh / all-zero image: `n_objects = 0`.
- `marching_cubes` failure for a label: that label is skipped (existing 
  behaviour) and does not contribute to the count.
- Single component: `n_objects = 1`, no behaviour change.
- **`n_labels != n_objects`:** the number of unique label values does not 
  equal the true object count when one or more labels contain multiple 
  disconnected components. `len(results)` reflects label count, not object 
  count. The proposed `n_objects` is the only way to recover the true number 
  of objects.

yajushikhurana:
## Summary

- Rename `center='centroid'` → `center='reference_centroid'` in `minkowski_tensors` to match C++ `--reference_centroid` flag (#73)
- Add `center='centroid_mesh'` to `minkowski_tensors` (volume-weighted center of mass = w010/w000); was previously only available in `minkowski_tensors_from_label_image` (#73)
- Add `center_scope='per_label'|'global'` parameter to both functions to control whether centroid is computed per label independently or from the full mesh (#73)
- Add `return_count=False` flag to both functions; when `True`, returns `(results, n_objects)` where `n_objects` is the total number of connected components (#80)
- Add three private helpers: `_compute_mesh_centroid`, `_ensure_outward_normals`, `_count_mesh_components`
- Add 15 new/updated tests covering all new functionality (80 tests total pass)

## Test plan

- [ ] Run `python -m pytest tests/test_api.py -v` — all 80 tests pass
- [ ] Verify `center='reference_centroid'` produces same results as old `center='centroid'`
- [ ] Verify `center='centroid_mesh'` shifts vectors near zero for a symmetric mesh
- [ ] Verify `return_count=True` returns `(dict, int)` tuple
- [ ] Verify `center_scope='global'` uses a single centroid across all labels

Closes #73
Closes #80

### DIFF
--- a/examples/pykarambola_demo.ipynb
+++ b/examples/pykarambola_demo.ipynb
@@ -108,7 +108,7 @@
    "outputs": [],
    "source": [
     "# Load a .poly file\n",
-    "tri = pykarambola.parse_poly_file('../test_suite/inputs/box_a=2_b=3_c=4.poly')\n",
+    "tri = pykarambola.parse_poly_file('../tests/fixtures/box_a=2_b=3_c=4.poly')\n",
     "\n",
     "print(f\"Vertices: {tri.n_vertices()}\")\n",
     "print(f\"Triangles: {tri.n_triangles()}\")"
@@ -221,7 +221,7 @@
    "source": [
     "# Load a mesh with labels\n",
     "tri_labelled = pykarambola.parse_poly_file(\n",
-    "    '../test_suite/inputs/two_boxes_sharing_vertex_with_labels.poly',\n",
+    "    '../tests/fixtures/two_boxes_sharing_vertex_with_labels.poly',\n",
     "    with_labels=True,\n",
     ")\n",
     "\n",
@@ -391,7 +391,7 @@
    "outputs": [],
    "source": [
     "# .off format\n",
-    "tri_off = pykarambola.parse_off_file('../test_suite/inputs/cuboid.off')\n",
+    "tri_off = pykarambola.parse_off_file('../tests/fixtures/cuboid.off')\n",
     "off_verts, off_faces, _ = arrays_from_triangulation(tri_off)\n",
     "print(f\"OFF: {tri_off.n_vertices()} vertices, {tri_off.n_triangles()} triangles\")\n",
     "\n",
@@ -407,7 +407,7 @@
    "outputs": [],
    "source": [
     "# .obj format\n",
-    "tri_obj = pykarambola.parse_obj_file('../test_suite/inputs/box.obj')\n",
+    "tri_obj = pykarambola.parse_obj_file('../tests/fixtures/box.obj')\n",
     "obj_verts, obj_faces, _ = arrays_from_triangulation(tri_obj)\n",
     "print(f\"OBJ: {tri_obj.n_vertices()} vertices, {tri_obj.n_triangles()} triangles\")\n",
     "\n",


yajushikhurana:
Fixed — updated all four path references in `pykarambola_demo.ipynb` from `../test_suite/inputs/` to `../tests/fixtures/` (commit `a082ce2`).

PR #83, opened 2026-03-10
Writing: Methods — architecture and acceleration strategy (#26, #27)

### LINKED ISSUE #26: Writing: Methods — architecture of pykarambola
Describe the overall structure of the Python port.

## Should cover
- Mapping from C++ karambola modules to Python modules
- Key design decisions: pure Python core, optional Cython acceleration, optional dependencies
- Supported input formats (poly, off, obj, glb) and which parsers are new vs ported

**Blocks:** #22, #37

### LINKED ISSUE #27: Writing: Methods — acceleration strategy (NumPy vectorization and Cython)
Describe the performance strategy without a dedicated benchmark figure.

## Should cover
- NumPy vectorization as the primary acceleration strategy
- Which parts of the computation resisted vectorization and required Cython
- Cython as an optional dependency (`pip install pykarambola[accel]`)
- Forward reference to runtime benchmark results in W10

**Blocks:** #22, #37

keisuke-ishihara:
## Summary

- Writes the **Architecture of pykarambola** subsection (closes #26): covers the mapping from C++ karambola to Python modules, replacement of GSL with NumPy's `eigh`, the pure-Python core with optional Cython and optional scikit-image/SciPy dependencies, and ported vs new file format parsers (`.poly`, `.off` ported; `.obj`, `.glb` new).
- Writes the **Acceleration strategy** subsection (closes #27): explains NumPy vectorization as the primary approach (vectorized precompute pass in `Triangulation`, `numpy.sum`/`numpy.einsum` in all compute functions), identifies the two graph-traversal routines that resisted vectorization (neighbor table, vertex-triangle CSR lookup), explains the optional Cython extension (`_accel.pyx`) with typed memoryviews, and notes the pure-Python fallback.
- Adds `harris2020numpy` and `virtanen2020scipy` to `references.bib` (both verified against Crossref).

Closes #26
Closes #27

## Test plan

- [ ] Compile manuscript PDF and confirm both subsections render without LaTeX errors
- [ ] Confirm citations `\cite{harris2020numpy}` and `\cite{virtanen2020scipy}` resolve in the bibliography

### DIFF
--- a/paper/manuscript.tex
+++ b/paper/manuscript.tex
@@ -82,7 +82,6 @@ Two initialisation steps resisted vectorization: building the edge-keyed triangl
 Both require sequential insertion into a hash map indexed by vertex-pair keys, a graph-traversal pattern with data-dependent branching that cannot be expressed as bulk array operations.
 These two routines are therefore provided as an optional Cython extension (\texttt{\_accel.pyx}) that uses typed memoryviews and eliminates interpreter overhead on the inner loops.
 When the extension is not installed, equivalent pure-Python implementations are used as a fallback.
-The impact of the Cython extension on wall-clock runtime is quantified in the Results section.
 
 \subsection*{New file format support (OBJ, GLB)}
 % TODO (issue #28)


keisuke-ishihara:
> **Architecture subsection** — "pykarambola retains support" (line starting the paragraph on format support): the `p` should be capitalised to "Pykarambola retains support" since it opens a new sentence.

fixed

> **Notation suggestion (non-blocking)** — `(F × 3) arrays, where F is the number of triangles`: the karambola paper uses F 2 as the set of triangulated faces, so we could write ( | F 2 | × 3 ) arrays to keep notation consistent with the reference. Not a required change.

Is 2 the F_2 mean it's a 2D surface? I think we can drop that notation here.
F is a scalar for the purpose of describing the shape of the array and it's bit confusing if there is a _2.

> **Question** — "The impact of the Cython extension on wall-clock runtime is quantified in the Results section." Does this mean the Results section will include a Cython-vs-pure-Python runtime comparison, in addition to the pykarambola-vs-C++ karambola benchmark?

The numpy vectorization was a big improvent for run time, but you didn't find much difference with or without C++, right? Which dataset was that on? It's possible that we would only see a difference with a very large mesh. I'm fine with not pursuing this specific benchmark, but what do we know so far?

keisuke-ishihara:
oh, i see. There is another set of notations where F: number of faces, V: vertices, and E: edges.
Euler char = V - E + F

i'm more familiar with this and am fine adopting F for number of faces.

keisuke-ishihara:
I deleted the sentence about Cython's impact on run time. Cython does make a difference for large meshes, but I don't think it's worth a sentence here. If you agree, you can approve the PR.

PR #107, opened 2026-04-07
Add SO(3)/O(3) invariants module for Minkowski tensors (ranks 0–2)

### LINKED ISSUE #102: Add SO(3)-invariant scalar construction from Minkowski tensors (ranks 0–2)
# Add SO(3)-invariant scalar construction from Minkowski tensors (Phase 1: ranks 0–2)

## Summary

Implement a modular pipeline that takes Minkowski tensors (ranks 0–2) as input and returns a vector of **basis invariants** — the irreducible SO(3)- or O(3)-invariant scalars obtained by contracting quantities that are not individually invariant (vectors and traceless matrices). Polynomial combinations of these basis invariants (e.g. products of scalars, scalar × dot-product, etc.) are **explicitly out of scope**: they carry no new geometric information beyond what the basis already encodes and can be constructed downstream via standard polynomial feature expansion if needed.

**Scope**: Ranks 0–2 only (14 standard tensors). Invariants from rank-3/4 tensors are out of scope and will be addressed in a separate follow-up issue.

## Motivation

Minkowski tensors decompose naturally into SO(3) irreducible representations (irreps) via the Clebsch-Gordan procedure:

- **Rank 0** (scalars) → `0e`
- **Rank 1** (vectors) → `1o`
- **Rank 2** (matrices) → `0e` (trace) ⊕ `2e` (traceless)

By constructing all **genuinely new** SO(3)-invariant contractions between irreps up to polynomial degree d=3, we obtain a compact basis of scalars that is:
- **Topology-independent** — defined via surface integrals, works for any orientable triangulated surface regardless of genus or convexity
- **Interpretable** — each invariant has direct geometric meaning tracing back to specific tensor interactions
- **Cross-field** — captures couplings between position, surface normal, and curvature distributions
- **Non-redundant** — scalar × scalar products and similar polynomial combinations are excluded; they are functions of existing basis elements, not new invariants
- **Chirality-sensitive** — SO(3)-only pseudo-scalars (triple vector determinants, commutator terms) detect handedness; O(3)-invariant features are blind to chirality, which matters for biological structures such as left- and right-handed helices and asymmetric cell shapes

This is distinct from and complementary to spherical harmonic descriptors, which provide high angular resolution for a single geometric field but require star-shaped topology assumptions.

The existing pykarambola classification pipeline uses eigenvalue-ratio features (β anisotropy indices) and achieves competitive balanced accuracy on biological shape datasets. The richer invariant set constructed here is expected to improve classification scores, particularly on datasets where eigenvalue-ratio features plateau, by capturing cross-field couplings and chirality that β indices cannot encode.

## What counts as a "basis invariant"

An invariant is a **basis invariant** if it cannot be expressed as a polynomial combination of lower-degree invariants already in the basis. The principle is:

- **Scalars are already invariant** — no contraction needed; include as-is at degree 1
- **Vectors are not individually invariant** — but their **dot products** v_i · v_j are genuinely new at degree 2
- **Traceless rank-2 matrices are not individually invariant** — but their **Frobenius inner products** Tr(A_i · A_j) are genuinely new at degree 2
- **Degree-3 contractions** are new only when they mix quantities across irrep types: quadratic forms v^T M v, triple matrix traces Tr(A B C), and (SO(3) only) pseudo-scalars from triple vector determinants and commutators

**Excluded as redundant polynomial features:**
- Scalar × scalar products (e.g. w100 × w200) — just a nonlinear function of existing scalars
- Scalar × (v_i · v_j) — just a scaling of an existing degree-2 invariant
- Scalar × Tr(A_i · A_j) — same reason

## Input tensors

The pipeline ingests the output of `minkowski_tensors(..., compute='standard')`:

- **Scalars** (4): w000, w100, w200, w300
- **Vectors** (4): w010, w110, w210, w310
- **Rank-2** (6): w020, w120, w220, w320, w102, w202

**Total: 14 tensors.** Rank-3/4 tensors (`w103`, `w104`, etc.) are explicitly out of scope.

## Architecture: Modular tensor registry

Invariants are constructed via a **modular registry system**:

```python
TENSOR_REGISTRY = {
    'w000': {'rank': 0, 'weight_type': 'position', 'irreps': ['0e']},
    'w100': {'rank': 0, 'weight_type': 'area', 'irreps': ['0e']},
    'w200': {'rank': 0, 'weight_type': 'curvature', 'irreps': ['0e']},
    'w300': {'rank': 0, 'weight_type': 'gaussian', 'irreps': ['0e']},
    'w010': {'rank': 1, 'weight_type': 'position', 'irreps': ['1o']},
    'w110': {'rank': 1, 'weight_type': 'area', 'irreps': ['1o']},
    'w210': {'rank': 1, 'weight_type': 'curvature', 'irreps': ['1o']},
    'w310': {'rank': 1, 'weight_type': 'gaussian', 'irreps': ['1o']},
    'w020': {'rank': 2, 'weight_type': 'position', 'irreps': ['0e', '2e']},
    'w120': {'rank': 2, 'weight_type': 'area', 'irreps': ['0e', '2e']},
    'w220': {'rank': 2, 'weight_type': 'curvature', 'irreps': ['0e', '2e']},
    'w320': {'rank': 2, 'weight_type': 'gaussian', 'irreps': ['0e', '2e']},
    'w102': {'rank': 2, 'weight_type': 'area_normal', 'irreps': ['0e', '2e']},
    'w202': {'rank': 2, 'weight_type': 'curvature_normal', 'irreps': ['0e', '2e']},
}

def compute_invariants(tensors_dict, max_degree=3, symmetry='SO3'):
    """
    Compute SO(3)- or O(3)-invariant basis scalars from Minkowski tensors.

    Returns the irreducible basis of invariants up to polynomial degree max_degree.
    Polynomial combinations of basis elements (e.g. products of scalars) are NOT
    included — use sklearn.preprocessing.PolynomialFeatures or similar downstream
    if you need an expanded feature vector for a linear model.

    Parameters
    ----------
    tensors_dict : dict
        Output from minkowski_tensors(..., compute='standard')
    max_degree : int, default=3
        Maximum polynomial degree for contractions
    symmetry : {'SO3', 'O3'}, default='SO3'
        - 'O3': Invariant under rotations and reflections (chirality-blind)
        - 'SO3': Invariant under rotations only; includes pseudo-scalars that
                 distinguish mirror-image shapes — relevant for chiral biological
                 structures

    Returns
    -------
    np.ndarray
        Basis invariant vector; length 155 (O3) or ~219 (SO3) at degree 3

    Notes
    -----
    This function performs pure algebraic contractions on whatever tensor values
    it receives. The choice of reference point is the caller's responsibility,
    made upstream via ``minkowski_tensors(center=...)``.

    Invariants involving translation-covariant tensors (w010–w310, w020–w320)
    are translation-covariant: different ``center`` choices produce different
    invariant vectors. Callers must use a consistent ``center`` across all
    shapes being compared.

    ``center='reference_centroid'`` must NOT be passed to
    ``minkowski_tensors()`` before calling this function. That mode applies a
    different shift to each rank-2 tensor while leaving rank-1 tensors at the
    origin, so the tensors live in inconsistent frames and their contractions
    are geometrically undefined.

    The translation-invariant subset — invariants built exclusively from
    w000, w100, w200, w300, w102, w202 — is unaffected by the center choice.

    For geometry-based analysis (size is informative): apply per-feature
    standardization across the dataset —
    ``sklearn.preprocessing.StandardScaler().fit_transform(X)`` where X has
    shape ``(n_objects, n_invariants)``.

    For shape analysis (scale-invariant): first normalize each invariant per
    object by an appropriate power of a chosen length scale (e.g.
    ``W_0**(1/3)`` from volume or ``W_1**(1/2)`` from surface area), then
    apply per-feature standardization across the dataset. The choice of
    length scale is application-dependent. The raw invariants must not be
    normalized per-object across invariants.
    """
```

### 1. Harmonic decomposition via Clebsch-Gordan projection

Extract irreducible components from all tensors. Following e3nn convention:

- **Rank-2 tensors**: Decompose into trace (`0e` scalar) + traceless (`2e` tensor)
  - Traceless component: `T_traceless = M - (Tr(M)/3)I`
  - Trace scalar: `t = Tr(M) / 3`

**Validation (following e3nn methodology):**
- Verify orthogonality: `T_traceless · T_trace = 0` (different irreps are orthogonal)
- Verify completeness: `T_traceless + (3t)I = M` (reconstruction recovers original)
- Verify norm preservation: `||M||_F² = ||T_traceless||_F² + 3t²` (energy conservation)

*References: Geiger & Smidt (2022) "e3nn: Euclidean Neural Networks"; Wigner-Eckart theorem and SO(3) harmonic analysis; Biedenharn & Louck for foundational representation theory*

### 2. SO(3)-invariant contractions via Wigner-3j symbols

Fully contract irreps of matching `l` and compatible parity. Valid invariant combinations are determined by **Wigner-3j selection rules**:

- **Selection rule**: `l_a ⊗ l_b → 0e` requires `l_a = l_b`, `parity_a × parity_b = +1`, and the triangle condition

All contractions computed via `numpy.einsum()`. Clebsch-Gordan coefficients are implicitly encoded in the einsum patterns.

#### Building blocks

**Scalars (8)** — already invariant, included as-is at degree 1:

| Symbol | Definition |
|---|---|
| s0 = w000 | volume |
| s1 = w100 | surface area |
| s2 = w200 | mean curvature integral |
| s3 = w300 | Gaussian curvature integral |
| s4 = Tr(w020)/3 | trace of position rank-2 tensor |
| s5 = Tr(w120)/3 | trace of area-weighted rank-2 tensor |
| s6 = Tr(w220)/3 | trace of curvature-weighted rank-2 tensor |
| s7 = Tr(w320)/3 | trace of Gaussian-weighted rank-2 tensor |

Note: `Tr(w102)/3 = w100` and `Tr(w202)/3 = w200` exactly — these are excluded to avoid redundancy. The traceless parts T4 = w102_tl and T5 = w202_tl are still included below.

**Vectors (4)** — not individually invariant:

| Symbol | Definition |
|---|---|
| v0 = w010 | centroid-like position vector |
| v1 = w110 | area-weighted centroid |
| v2 = w210 | curvature-weighted centroid |
| v3 = w310 | Gaussian-weighted centroid |

**Traceless rank-2 tensors (6)** — not individually invariant; each defined as M_tl = M − (Tr(M)/3)I:

| Symbol | Definition |
|---|---|
| T0 = w020_tl | traceless part of w020 |
| T1 = w120_tl | traceless part of w120 |
| T2 = w220_tl | traceless part of w220 |
| T3 = w320_tl | traceless part of w320 |
| T4 = w102_tl | traceless part of w102 |
| T5 = w202_tl | traceless part of w202 |

#### Degree-2 contractions (31 total)

**Vector dot products** `v_i · v_j` for all i ≤ j (10 invariants):

```
v0·v0,  v0·v1,  v0·v2,  v0·v3
        v1·v1,  v1·v2,  v1·v3
                v2·v2,  v2·v3
                        v3·v3
```

**Frobenius inner products** `Tr(T_i · T_j)` for all i ≤ j (21 invariants):

```
T0:T0,  T0:T1,  T0:T2,  T0:T3,  T0:T4,  T0:T5
        T1:T1,  T1:T2,  T1:T3,  T1:T4,  T1:T5
                T2:T2,  T2:T3,  T2:T4,  T2:T5
                        T3:T3,  T3:T4,  T3:T5
                                T4:T4,  T4:T5
                                        T5:T5
```

#### Degree-3 contractions (~116 O3, ~180 SO3)

**Quadratic forms** `v_i^T T_k v_j` for all k ∈ {T0..T5}, i ≤ j (60 invariants):

For each of the 6 traceless matrices T_k, and each of the 10 symmetric vector pairs (v_i, v_j) with i ≤ j. Note: `v_i^T T_k v_j = v_j^T T_k v_i` since T_k is symmetric, so the i ≤ j constraint avoids double-counting.

```
T_k = T0:  v0ᵀT0v0, v0ᵀT0v1, v0ᵀT0v2, v0ᵀT0v3, v1ᵀT0v1, v1ᵀT0v2, v1ᵀT0v3, v2ᵀT0v2, v2ᵀT0v3, v3ᵀT0v3
T_k = T1:  (same 10 pairs)
...
T_k = T5:  (same 10 pairs)
```

**Triple matrix traces** `Tr(T_i T_j T_k)` under cyclic+reversal symmetry (~56 invariants, exact count pending Open Q#2):

Symmetries: cyclic + reversal generate the full S₃ (≅ D₃, order 6). By Burnside's lemma the number of distinct invariants is C(n+2,3) = n(n+1)(n+2)/6. For n=6: C(8,3) = **56 exactly**.

**Triple vector determinants** `det([v_i, v_j, v_k])` for all i < j < k (SO(3) only, 4 invariants):

```
det([v0,v1,v2]),  det([v0,v1,v3]),  det([v0,v2,v3]),  det([v1,v2,v3])
```

**Commutator pseudo-scalars** `ε_pqr [T_i, T_j]_pq (v_k)_r` for all i < j, all k (SO(3) only, up to 60 invariants, pending Open Q#3):

For each of the C(6,2) = 15 off-diagonal matrix pairs (T_i, T_j) with i < j, and each of the 4 vectors v_k. Note: [T_i, T_j] = T_i T_j − T_j T_i is antisymmetric, so ε_pqr [T_i, T_j]_pq is well-defined and changes sign under reflection.

```
[T0,T1]·v0,  [T0,T1]·v1,  [T0,T1]·v2,  [T0,T1]·v3
[T0,T2]·v0,  [T0,T2]·v1,  [T0,T2]·v2,  [T0,T2]·v3
...
[T4,T5]·v0,  [T4,T5]·v1,  [T4,T5]·v2,  [T4,T5]·v3
```

#### Summary table

| Degree | Type | Count (O3) | Count (SO3) |
|---|---|---|---|
| 1 | Scalars | 8 | 8 |
| 2 | Vector dot products | 10 | 10 |
| 2 | Frobenius inner products | 21 | 21 |
| 3 | Quadratic forms | 60 | 60 |
| 3 | Triple matrix traces | 56 | 56 |
| 3 | Triple vector determinants | 0 | 4 |
| 3 | Commutator pseudo-scalars | 0 | ~60 |
| **Total** | | **155** | **~219** |

> Note: SO3 total is approximate because the exact count of linearly independent commutator pseudo-scalars is to be confirmed by a numerical rank test during implementation (see ~~Q#3~~ below).
---

## Pre-implementation open questions

These must be resolved before finalising the implementation. Each maps to a verification item in the acceptance criteria.

### ~~Open question #1~~ — Linear dependencies among `0e` scalars ✓ RESOLVED

**Decision: Option A** — `Tr(w102)` and `Tr(w202)` are excluded from the scalar pool.

The identities `Tr(w102) = w100` and `Tr(w202) = w200` are exact by construction:
- `Tr(w102)` = ∫_S **n**·**n** dA = ∫_S dA = `w100` (surface area)
- `Tr(w202)` = ∫_S H (**n**·**n**) dA = ∫_S H dA = `w200` (mean curvature integral)

Including them would introduce known rank-2 redundancy into the scalar pool. Since the identities are analytical (not approximate), there is no value in retaining them; any downstream model that requires the redundancy can reconstruct them from s1 and s2.

**Remaining scalars s4–s7** (`Tr(w020)/3` through `Tr(w320)/3`) are second-moment-type integrals of position squared over the domain. They are not expressible in terms of w000–w300 in general (those are degree 1 in position; the traces of rank-2 position tensors are degree 2). No other exact dependencies exist among s0–s7.

Note: the Schröder-Turk (2013) eq. 10 relation `Q W_ν = ν W^{0,2}_ν + (3−ν) W^{1,1}_{ν+1}` does **not** introduce dependencies among the standard tensors. The W^{1,1}_ν tensors (w111, w211, w311 — asymmetric rank-2 tensors integrating **x**⊗**n**) are outside the standard set and not computed by `minkowski_tensors()`.

**Numerical rank test** (to be run during implementation): compute the full O3 feature matrix across a diverse shape set (sphere, cube, ellipsoid, torus, random convex polytope) and verify that the numerical rank equals 155. If any pair of traceless tensors T_i, T_j satisfies T_i = α T_j due to a geometric identity, Frobenius products Tr(T_i · T_k) would be linearly dependent on Tr(T_j · T_k); resolve analytically before finalising.

### ~~Open question #2~~ — Exact count for triple rank-2 contractions ✓ RESOLVED

The contraction `Tr(T_i T_j T_k)` is symmetric under:
- **Cyclic shifts**: `Tr(T_i T_j T_k) = Tr(T_j T_k T_i) = Tr(T_k T_i T_j)` — trace cyclicity
- **Reversal**: `Tr(T_i T_j T_k) = Tr(T_k T_j T_i)` — since each T is symmetric, `(ABC)ᵀ = CᵀBᵀAᵀ = CBA`

These two generators produce the **full symmetric group S₃** acting on (i,j,k) — not a proper subgroup — because the dihedral group D₃ and S₃ are isomorphic (both have order 6). The symmetry group is therefore all 6 permutations of the three indices.

#### Burnside's lemma for n traceless symmetric rank-2 tensors

| Group element | Action on (i,j,k) | Fixed triples | Count |
|---|---|---|---|
| e (identity) | (i,j,k) | all triples | n³ |
| c (cyclic) | (j,k,i) | i=j=k | n |
| c² (cyclic²) | (k,i,j) | i=j=k | n |
| r (reversal) | (k,j,i) | i=k | n² |
| rc | (i,k,j) | j=k | n² |
| rc² | (j,i,k) | i=j | n² |

$$\text{distinct orbits} = \frac{1}{6}\left(n^3 + n + n + n^2 + n^2 + n^2\right) = \frac{n(n+1)(n+2)}{6} = \binom{n+2}{3}$$

This is the number of **multisets of size 3** from n elements — consistent with the full S₃ symmetry.

**For n = 6 traceless tensors**: $\binom{8}{3} = \mathbf{56}$ distinct invariants (exact, not approximate).

### ~~Open question #3~~ — SO(3)-only pseudo-scalars from rank-2 commutators ✓ RESOLVED

**Decision**: Include all 60 commutator pseudo-scalars in `symmetry='SO3'`; final count to be confirmed by numerical rank test.

For any two symmetric rank-2 tensors A, B and vector v, the quantity `ψ(A, B, v) = ε_ijk [A, B]_ij v_k` is a valid SO(3) pseudo-scalar that changes sign under reflection. It equals the dot product of v with the axial vector of the commutator [A, B] = AB − BA. Since A = B gives [A, A] = 0, only the 15 off-diagonal pairs (i < j) contribute.

The 64 pseudo-scalars (60 commutator terms + 4 triple-vector determinants) are structurally distinct: triple-vector determinants are degree 3 in vectors, while commutator terms are degree 2 in rank-2 tensors × degree 1 in vectors. They are not linearly dependent in general, but geometric coincidences on specific shape families (e.g. shapes where all T_i are proportional, so all commutators vanish) may reduce the observed rank.

**Numerical rank test** (to be run during implementation): compute all 64 pseudo-scalars on a diverse shape set (sphere, cube, ellipsoid, torus, random convex polytope) and check the rank of the resulting 64-column matrix. If the rank is substantially below 64, identify which terms are redundant analytically and remove them from the SO3 feature set before finalising the invariant count. Update the summary table accordingly.

### ~~Open question #4~~ — Centroid dependence in cross-contractions ✓ RESOLVED

`compute_invariants()` performs pure algebraic contractions on pre-computed tensors and has no centroid logic. The reference point is chosen upstream via `minkowski_tensors(center=...)` and is already baked into the tensor values by the time they are ingested.

The invariant vector is intentionally centroid-dependent for invariants involving position tensors. `center='reference_centroid'` must not be used: it applies a different shift to each rank-2 tensor while leaving rank-1 tensors at the origin, producing tensors in inconsistent frames whose contractions are geometrically undefined. The valid options are `center=None`, `center='centroid_mesh'`, or an explicit `(3,)` array — all of which shift vertices uniformly before any tensor is computed.

The translation-invariant subset consists of invariants built exclusively from w000, w100, w200, w300, w102, w202:
- s0–s3 (4 of the 8 scalars; Tr(w102) and Tr(w202) are excluded as exact duplicates of s1, s2)
- T4:T4, T4:T5, T5:T5 (3 Frobenius products)
- Tr(T4 T4 T4), Tr(T4 T4 T5), Tr(T4 T5 T5), Tr(T5 T5 T5) (4 triple traces)

This is documented in the `compute_invariants()` docstring.

---

## Implementation notes

- All contractions are pure `numpy.einsum()` operations — no external dependencies
- Input tensors must be **fully symmetrized** over all index permutations (guaranteed by `minkowski_tensors()`)
- `compute_invariant_labels()` must produce a **deterministic**, run-stable ordering of labels — this is a hard requirement for ML reproducibility

### Code organization

```
pykarambola/invariants.py (new module)
├── TENSOR_REGISTRY
├── traceless_rank2(M) → (3,3)
├── trace_rank2(M) → float
├── compute_invariants(tensors_dict, max_degree=3, symmetry='SO3') → np.ndarray
├── compute_invariant_labels(symmetry='SO3') → list[str]
└── _enumerate_invariant_contractions(symmetry='SO3')

tests/test_invariants.py
├── test_harmonic_decomposition()
│   ├── verify orthogonality of rank-2 irrep decomposition
│   ├── verify trace + traceless recovers original
│   └── verify norm preservation
├── test_rotational_invariance()
│   ├── apply 100+ random SO(3) rotations
│   ├── verify relative deviation < 1e-8
│   └── report max deviation across all invariants
├── test_scalar_linear_dependencies()        ← NEW (Open Q#1)
│   ├── verify Tr(w102) == w100
│   ├── verify Tr(w202) == w200
│   └── check for any other exact dependencies
├── test_clebsch_gordan_consistency()
└── test_invariant_reproducibility()
```

## Validation: Rotational Invariance Test Suite

```python
def test_rotational_invariance(n_trials=100, tol=1e-8, symmetry='SO3'):
    from scipy.spatial.transform import Rotation

    tensors_original = minkowski_tensors(verts_reference, faces_reference, compute='standard')
    invariants_original = compute_invariants(tensors_original, symmetry=symmetry)

    max_deviation = 0.0
    for trial in range(n_trials):
        R = Rotation.random().as_matrix()

        tensors_rotated = {
            'w000': tensors_original['w000'],
            'w100': tensors_original['w100'],
            'w200': tensors_original['w200'],
            'w300': tensors_original['w300'],
            'w010': R @ tensors_original['w010'],
            'w110': R @ tensors_original['w110'],
            'w210': R @ tensors_original['w210'],
            'w310': R @ tensors_original['w310'],
            'w020': np.einsum('ia,jb,ab->ij', R, R, tensors_original['w020']),
            'w120': np.einsum('ia,jb,ab->ij', R, R, tensors_original['w120']),
            'w220': np.einsum('ia,jb,ab->ij', R, R, tensors_original['w220']),
            'w320': np.einsum('ia,jb,ab->ij', R, R, tensors_original['w320']),
            'w102': np.einsum('ia,jb,ab->ij', R, R, tensors_original['w102']),
            'w202': np.einsum('ia,jb,ab->ij', R, R, tensors_original['w202']),
        }

        invariants_rotated = compute_invariants(tensors_rotated, symmetry=symmetry)
        relative_error = np.linalg.norm(invariants_rotated - invariants_original) / (
            np.linalg.norm(invariants_original) + 1e-14
        )
        max_deviation = max(max_deviation, relative_error)

    assert max_deviation < tol, f"Max deviation {max_deviation:.2e} exceeds tolerance {tol}"
    return True, max_deviation
```

Use multiple reference meshes: unit cube, icosphere, random convex polytope, non-convex shape (torus).

## Acceptance criteria

**Core implementation:**
- [ ] `traceless_rank2(M)` implemented: extracts `2e` irrep
- [ ] `trace_rank2(M)` implemented: extracts `0e` scalar component
- [ ] Harmonic decomposition validation tests pass: orthogonality, completeness, norm preservation
- [ ] `compute_invariants(tensors_dict, max_degree=3, symmetry='SO3')` returns `np.ndarray`
- [ ] `compute_invariant_labels(symmetry='SO3')` returns matching list of human-readable labels with **deterministic ordering**
- [ ] `TENSOR_REGISTRY` covers the 14 standard tensors (ranks 0–2 only)
- [ ] Clebsch-Gordan consistency test passes: all contractions satisfy Wigner-3j selection rules
- [ ] **Rotational invariance test passes** for both SO(3) and O(3): `max_relative_deviation < 1e-8` across 100+ random SO(3) rotations on multiple mesh types
- [ ] Code references Geiger & Smidt (2022) and adopts e3nn irrep notation ("0e", "1o", "2e")
- [ ] Benchmark documents invariant count, runtime, and invariance deviation for both symmetry groups
- [ ] API docstring explicitly states that polynomial combinations of basis invariants are out of scope

- [ ] `CHANGELOG.md` updated under `[Unreleased]` with entry for `compute_invariants()` and `compute_invariant_labels()`

**Pre-implementation open questions (must be resolved before merging):**
- [x] **Open Q#1 resolved**: Option A chosen — `Tr(w102)` and `Tr(w202)` excluded from scalar pool; 8 independent scalars; s4–s7 verified to be independent of s0–s3 analytically; numerical rank test to confirm full-feature-matrix rank = 155 during implementation
- [x] **Open Q#2 resolved**: Exact count = C(n+2,3) = n(n+1)(n+2)/6 by Burnside's lemma (D₃ ≅ S₃); for n=6 gives 56 exactly
- [x] **Open Q#3 resolved**: All 60 commutator pseudo-scalars included in `symmetry='SO3'`; final count subject to numerical rank test on all 64 pseudo-scalars during implementation
- [x] **Open Q#4 resolved**: `compute_invariants()` has no center parameter; `center='reference_centroid'` documented as invalid input; translation-covariant vs. translation-invariant subsets identified; warning added to `compute_invariants()` docstring

## Dependencies and references

**Papers & Books:**
- Geiger & Smidt (2022). "e3nn: Euclidean Neural Networks Library" — SO(3) irrep theory, Clebsch-Gordan decomposition, and validation methodology
- Mecke & Schroder-Turk. "Minkowski Tensor Shape Analysis" — integral geometry framework
- Kazhdan et al. (2003). "Rotation Invariant Spherical Harmonic Representation of 3D Shape Descriptors" — comparison baseline
- Biedenharn & Louck. *The Racah-Wigner Algebra in Quantum Theory* — foundational SO(3) representation theory and Wigner-3j symbols
- NIST DLMF — reference for Wigner coefficients and CG decomposition

**Blocking:**
- **Blocks**: Future issues on shape classification or descriptor applications using these invariants
- **Blocked by**: None
- **Follow-up**: Rank-3/4 invariant construction (requires #101) will be a separate issue
- **Milestone**: M0

keisuke-ishihara:
## Summary

Implements `pykarambola/invariants.py` and a comprehensive test suite for constructing SO(3)- and O(3)-invariant basis scalars from Minkowski tensors (ranks 0–2).

## Implementation strategy

The original issue plan specified a fixed 14-tensor registry. Based on follow-up discussion in the issue thread, the implementation was updated to support **arbitrary tensor subsets** — ranks are inferred dynamically from tensor shapes, and invariant enumeration adapts to whatever tensors are provided:

- **No fixed `TENSOR_REGISTRY`** — rank inferred from tensor shape: scalar → rank 0, `(3,)` → rank 1, `(3,3)` → rank 2
- **`decompose_all(tensors_dict)`** accepts any dict of tensor name → value with no hardcoded names
- **Dynamic enumeration** — invariant counts depend on how many scalars/vectors/matrices are present; all pairs/multisets are generated at runtime via `itertools`
- **Conditional scalar deduplication** — the known identities `Tr(w102)/3 = w100` and `Tr(w202)/3 = w200` are only applied when *both* the dependent and source tensors are present in the input
- **`compute_invariants()` returns `dict[str, float]`** (not a plain `np.ndarray`) with deterministic, human-readable label keys — designed for reproducible ML feature extraction
- **`deduplicate_scalars` parameter** controls whether the w102/w202 trace redundancy is removed

## What's included

- `pykarambola/invariants.py`: `_infer_rank`, `decompose_all`, `_degree1_scalars`, `_vector_dot_products`, `_frobenius_inner_products`, `_quadratic_forms`, `_triple_traces`, `_triple_vector_dets`, `_commutator_pseudoscalars`, `compute_invariants`, `compute_invariant_labels`
- `tests/test_invariants.py`: harmonic decomposition (orthogonality, completeness, norm preservation), rotational invariance (100+ random SO(3) rotations, tol < 1e-8), translation invariance on the translation-invariant subset, and Clebsch-Gordan consistency
- `CHANGELOG.md` updated under `[Unreleased]`

## Invariant counts (with full 14-tensor input)

| Degree | Type | O(3) | SO(3) |
|---|---|---|---|
| 1 | Scalars (after dedup) | 8 | 8 |
| 2 | Vector dot products | 10 | 10 |
| 2 | Frobenius inner products | 21 | 21 |
| 3 | Quadratic forms | 60 | 60 |
| 3 | Triple matrix traces | 56 | 56 |
| 3 | Triple vector determinants | 0 | 4 |
| 3 | Commutator pseudo-scalars | 0 | ~60 |
| **Total** | | **155** | **~219** |

## Test plan

- [ ] `pytest tests/test_invariants.py` passes
- [ ] Rotational invariance holds for both `symmetry='SO3'` and `symmetry='O3'` on sphere, cube, ellipsoid, torus meshes
- [ ] Translation invariance verified for the translation-invariant subset (w000, w100, w200, w300, w102, w202 only)
- [ ] Conditional deduplication tested: w102/w202 trace scalars removed when w100/w200 present, retained when absent
- [ ] Partial tensor dict input (e.g. only w000, w010, w020) produces consistent counts and labels
- [ ] `compute_invariant_labels()` output length matches `compute_invariants()` dict length for all parameter combinations

Closes #102


keisuke-ishihara:
## Code Review

Reviewed against both the original issue plan and the modified flexible-input strategy (second issue comment). One item from my earlier review is retracted; the remaining issues are below.

---

### Retraction

~~`compute_invariant_labels` is missing~~ — the modified strategy changed `compute_invariants` to return `dict[str, float]`, making a separate labels function unnecessary. The PR description's test-plan item mentioning it is stale but not a code bug.

---

### Bugs

**1. Sign error in `_commutator_pseudoscalars` (invariants.py:297)**

The issue thread specifies the contraction as `ε_{ijk} [T_a, T_b]_{ij} v_k`. Expanding for an antisymmetric commutator C gives `2 * [C[1,2], C[2,0], C[0,1]] · v`. The code computes:

```python
axial = np.array([comm[2, 1], comm[0, 2], comm[1, 0]])
```

Since `comm[2,1] = -comm[1,2]` (antisymmetry), this gives the wrong sign and is off by a factor of 2 from the specified formula. The inline comment is the source of confusion. Tests don't catch this because they only verify sign-flipping under reflection, which holds regardless of an overall sign.

**2. Non-symmetric rank-2 inputs corrupt the `2e` component**

`_traceless_rank2(M)` computes `M - Tr(M)/3 * I` without symmetrizing first. For a non-symmetric M this contaminates the `2e` component with the antisymmetric (`l=1`) part. The docstring says _"traceless symmetric part"_ but the code does not enforce symmetry. The PR claims to accept "arbitrary tensors" — this contract is silently broken for non-symmetric inputs.

---

### Missing acceptance-gate items (from the issue thread)

**M1** — acceptance gate requires "all three decomposition properties hold on every test mesh (unit cube, icosphere, random convex hull)." Tests use only hand-crafted synthetic tensors; no actual mesh-derived Minkowski tensors are tested.

**M2** — acceptance gate requires "scan for any other near-dependencies among the 10 scalars (compute rank of the 10×N matrix across N meshes; should be 8)." Only the two known w102/w202 identities are tested; no general rank check.

**M5** — "Filter zeros: compute commutators on several meshes and remove any that are numerically zero." All `C(m,2) × n` commutator terms are generated unconditionally. For the unit_cube_tensors fixture (all rank-2 tensors are diagonal multiples of the identity), many commutators are identically zero — not filtered or documented.

**M7** — Performance benchmarks (1K/10K/100K triangles, breakdown by degree) and bitwise reproducibility test are both absent.

---

### `__init__.py` not updated

`pykarambola/invariants.py` is added but `__init__.py` is not updated. The CHANGELOG advertises `pykarambola.invariants` as a public module, but `from pykarambola import compute_invariants` fails. At minimum `compute_invariants` and `decompose_all` should be re-exported.

---

### Non-standard irrep label `1e` for polar vectors

Both the issue thread and PR use `'1e'` for rank-1 tensors. In e3nn / physics convention (per the cited Geiger & Smidt 2022 reference): `1e` = axial/pseudovector (even parity under inversion), whereas Minkowski tensors like `w010` are polar vectors (odd parity = `1o`). This does not corrupt computed values since parity is handled by the O3/SO3 switch, not the label — but it contradicts the cited reference and will confuse users familiar with e3nn.

---

### Minor

- **`~60` in the PR summary table** should be exactly `60` (C(6,2) × 4 vectors = 60).
- **Non-seeded `np.random.randn`** in `test_count_four_vectors`, `test_count_three_matrices`, `test_count_six_matrices` — inconsistent with the rest of the suite which uses `np.random.default_rng(seed)`.
- **Tautological CG tests** — `test_all_invariant_types_satisfy_selection_rules` has every branch as `pass`; the other CG tests verify basic NumPy operations, not the invariants module itself.
- **"Complete basis" claim** — syzygies among the listed invariants (e.g. Cayley-Hamilton relations among the triple-trace terms) are not addressed.

keisuke-ishihara:
## Code Review — `fefca63` (Add `symmetry='SO2'` to `compute_invariants`)

---

### Bugs

**1. Sign error in `tp_im` (`invariants.py:438`)**

The docstring says Im = `Im[ conj(c) · (a·b) ]`, but the code computes the **negative** of that:

```python
im_val = re_ab * c[1] - im_ab * c[0]   # actually = -Im[conj(c)*(a·b)]
```

Verified numerically:
```
True  Im[conj(c)*(a*b)]: 80.0
Code  im_val:            -80.0
```

The `tp_im` label is therefore a misnomer for all non-zero values. None of the tests catch this because the known-value test chooses inputs that make `im_val = 0`, and the invariance tests are indifferent to sign.

The inline comment on the same line compounds the confusion: `# Re(conj(b)*a) ... but we want a·b as complex`. `Re(conj(b)*a)` is the real dot product `ax*bx + ay*by`, whereas the code computes `ax*bx - ay*by = Re(a*b)`. These are different quantities; the comment is wrong on both counts.

**2. Silent key collision in `_so2_degree1_scalars` (`invariants.py:344–365`)**

Keys `{name}_z` (from rank-1 tensors) and `{name}_zz` (from rank-2 tensors) silently overwrite any rank-0 scalar whose name happens to end in `_z` or `_zz`:

```python
tensors = {'v_z': 99.0, 'v': np.array([1., 2., 3.])}
scalars = _so2_degree1_scalars(_decompose_so2(tensors))
scalars['v_z']  # → 3.0  (correct: 99.0 is silently lost)
```

The function is documented as accepting arbitrary tensor names; this invariant is violated. No test covers it.

**3. Invalid `symmetry` string falls through silently (`invariants.py:533`)**

A typo or wrong case (e.g. `symmetry='so2'`, `symmetry='SO2 '`) skips the SO(2) branch and silently produces SO(3) results. The `Literal` annotation is documentation-only at runtime:

```python
compute_invariants({'w010': np.array([1.,0.,0.])}, symmetry='so2')
# → {'dot_w010_w010': 1.0}   (SO(3) output, no error)
```

---

### Missing items

**4. Numerical rank test not performed**

The plan explicitly states: *"A numerical rank test is recommended during implementation."* None was added. With 660 degree-3 triple products for 14 tensors, there has been no verification that the set is actually linearly independent — or if not, which invariants are redundant. For `tp_im` self-pairs (`a=b`), the relationship to `tp_re` self-pairs across different `c` choices deserves scrutiny.

**5. No tests with mesh-derived Minkowski tensors**

All SO(2) invariance tests use synthetic arrays. Unlike the SO(3) tests (which at least use `unit_cube_tensors`), the SO(2) suite has no test exercising the full `mesh → minkowski_tensors → compute_invariants(symmetry='SO2')` pipeline.

---

### Minor

**6. Unused fixture `all_rank_tensors`**

Defined in `TestSO2Invariants` but not used by any test method.

**7. Redundant variable `lbl` in `_so2_collect_doublets` (`invariants.py:379`)**

`lbl` in the comprehension is always equal to the argument `label` (by the filter). Using `f"{name}_{label}"` directly would be clearer:

```python
# current:  (f"{name}_{lbl}", val)
# clearer:  (f"{name}_{label}", val)
```

pykarambola/invariants.py
```diff
@@ -0,0 +1,615 @@
...
+    if isinstance(tensor, (int, float)):
+        return 0
+    arr = np.asarray(tensor)
+    if arr.ndim == 0:
+        return 0
+    if arr.shape == (3,):
+        return 1
+    if arr.shape == (3, 3):
+        return 2
+    raise ValueError(f"Unsupported tensor shape: {arr.shape}. Expected scalar, (3,), or (3,3).")
```
keisuke-ishihara, line 161: **High: default usage silently produces non-invariant features.**

`minkowski_tensors()` defaults to `compute_eigensystems=True` (api.py:215), so the natural call `compute_invariants(minkowski_tensors(verts, faces))` feeds eigensystem keys in here. `_infer_rank` classifies `*_eigvals` (shape `(3,)`) as a rank-1 vector and `*_eigvecs` (shape `(3,3)`) as a rank-2 tensor. Eigenvector matrices are orthogonal, not spherical tensors; `_traceless_rank2` symmetrizes them and the result does not transform as a proper l=2 irrep.

Verified on a 3x2x1 box under a random rotation: **995 of 1171 returned "invariants" are not rotation-invariant.** The existing `test_eigensystem_keys_inflate_count` documents this as expected rather than guarding against it.

Suggestion: validate that a `(3,3)` input is symmetric (within tol) before treating it as rank-2 — that alone turns the dangerous eigvec case into an error while leaving genuine Minkowski tensors untouched. Alternatively skip/warn on known non-tensor key suffixes (`_eigvals`, `_eigvecs`).

pykarambola/invariants.py
```diff
@@ -0,0 +1,615 @@
...
+    result = {}
+
+    # |m|=1 doublets: both 'xy' (rank-1) and 'xz' (rank-2)
+    m1_doublets = sorted(
+        _so2_collect_doublets(so2_dec, 'xy') + _so2_collect_doublets(so2_dec, 'xz'),
+        key=lambda x: x[0],
+    )
+    for i, (ci, di) in enumerate(m1_doublets):
+        for cj, dj in m1_doublets[i:]:
+            result[f"d1_{ci}_{cj}"] = float(np.dot(di, dj))
```
keisuke-ishihara: The dot product keeps only `Re(a·conj(b))`. The antisymmetric part `Im(a·conj(b)) = a_x b_y − a_y b_x` is also SO(2)-invariant and independent for distinct doublets (and parity-even under z→−z, so it would survive in O(2) too). Degree-3 (`tp_re`/`tp_im`) does keep both parts, so this is a slight asymmetry. If the omission is intentional, consider softening the "complete basis of polynomial invariants" phrasing in the `compute_invariants` docstring.

pykarambola/spharm_invariants.py
```diff
@@ -0,0 +1,206 @@
...
+
+    for idx, (l1, l2, l) in enumerate(triples):
+        # Precompute nonzero CG entries for this triple
+        cg_entries = []
+        for m1 in range(-l1, l1 + 1):
+            for m2 in range(-l2, l2 + 1):
+                m = m1 + m2
+                if abs(m) > l:
+                    continue
+                cg_val = _cg(l1, m1, l2, m2, l, m)
```
keisuke-ishihara: The compat tests (`test_shinvariants_compat.py`) are `importorskip`-gated on `aicsshparam`, which isn't installed in CI, so this local bispectrum path and the `_HAS_UPSTREAM` delegation branch are both unexercised. The compat test docstring notes local vs upstream disagree for odd `l1+l2+l` triples. Consider a `# TODO(#137)` here so the known odd-J discrepancy is visible at the code site, not just in the test docstring.

pyproject.toml
```diff
@@ -44,6 +44,8 @@ accel = ["cython>=3.0"]
 glb = ["trimesh"]
 stl = ["numpy-stl"]
 notebooks = ["scikit-image>=0.19", "matplotlib", "seaborn", "pandas", "medmnist", "tifffile", "scikit-learn"]
+# aicsshparam is not yet on PyPI — uncomment and pin the version once released (issue #137)
```
keisuke-ishihara: This `aicsshparam` comment sits directly above the `benchmark` extra, but `benchmark` doesn't reference aicsshparam — the comment looks orphaned/misplaced relative to what it annotates.

keisuke-ishihara (commented):
Review of the invariants module. Ran the suite (187 passed) and verified the SO(3)/O(3)/SO(2)/O(2) math by hand and empirically — core math is correct and the mesh-level rotation-invariance tests are excellent. One high-priority footgun in the default usage, plus a few smaller notes left inline.

pykarambola/invariants.py
keisuke-ishihara: Fixed in 021e729. `compute_invariants` now strips `*_eigvals`/`*_eigvecs` keys on input with a `UserWarning` (`_drop_non_tensor_keys`), so the natural `compute_invariants(minkowski_tensors(verts, faces))` call is safe.

I went with a name-based filter rather than the symmetry check I suggested above: a `(3,3)` symmetry check would catch `_eigvecs` but not `_eigvals` (a legit `(3,)` shape that's a scalar triple, not a geometric vector), and the rank-2 path deliberately symmetrizes inputs (bb54394), so a hard symmetry-raise would fight that design.

Verified on the 3×2×1 box under a random rotation: was 995/1171 non-invariant → now 155 invariants, 0 non-invariant (matches the documented O(3) count). Replaced the two tests that documented the footgun with tests that confirm the full-output result equals the `compute_eigensystems=False` result and stays rotation-invariant; added `TestDropNonTensorKeys`. Full suite: 191 passed.

pykarambola/invariants.py
keisuke-ishihara: Addressed in 4de34b7. The antisymmetric part `Im(a·conj(b)) = a_x b_y − a_y b_x` (the 2D wedge) is now emitted alongside the dot product, as `x1_` (|m|=1) and `x2_` (|m|=2) keys for distinct doublet pairs (it vanishes on the diagonal). The wedge shares the same z-reflection parity as the dot product, so the O(2) filter handles `x1_` exactly like `d1_`.

Standard 14-tensor SO(2) count: 754 → 814 (+45 `x1_`, +15 `x2_`). Verified the new terms are invariant under z-rotation, change under x-rotation (so they're non-trivial), and that O(2) drops the mixed-parity (`_xy`×`_xz`) wedges. Full suite: 192 passed.

pykarambola/spharm_invariants.py
keisuke-ishihara: Resolved in 816db6e by removing the feature entirely. Per #137 the SPHARM invariants are being contributed upstream to aics-shparam, and shipping a shim against an unreleased package (with the local odd-J bispectrum incompleteness noted here) isn't worth it. Deleted `spharm_invariants.py` and `test_shinvariants_compat.py`, plus the CHANGELOG entry and the stale aicsshparam comment in `pyproject.toml`. This also keeps the PR scoped to #102.

(Core spherical Minkowski tensors in `spherical.py`/`test_wigner3j.py` are a separate feature and untouched.)

pyproject.toml
keisuke-ishihara: Removed in 816db6e along with the rest of the SPHARM/aicsshparam code (#137). The `benchmark` extra stays (it's ML deps for the #102 invariants benchmark, not SPHARM-related).

PR #108, opened 2026-04-19
Close #90, #94, #99: Triangulation input, surface checks, label image padding (v0.3.0)

### LINKED ISSUE #90: Triangulation objects returned by parsers cannot be passed directly to minkowski_tensors
## Problem

All four file parsers (`parse_obj_file`, `parse_off_file`, `parse_poly_file`, `parse_glb_file`) return a `Triangulation` object. However, `minkowski_tensors` only accepts raw numpy arrays:

```python
def minkowski_tensors(verts, faces, labels=None, ...):
    verts = np.asarray(verts, dtype=np.float64)
    faces = np.asarray(faces, dtype=np.int64)
```

This means the following natural usage pattern fails:

```python
tri = parse_obj_file("model.obj")
result = minkowski_tensors(tri, ...)  # TypeError / wrong shape
```

## Why `tri._verts` cannot be used directly either

The parsers build a `Triangulation` in **append-mode**: vertices and faces are accumulated in Python lists (`_verts_list`, `_faces_list`) via `append_vertex()` / `append_triangle()`. The numpy arrays `_verts` and `_faces` are `None` until `_consolidate_lists()` is explicitly called:

```python
def __init__(self):
    self._verts_list = []
    self._faces_list = []
    self._verts = None   # not populated yet
    self._faces = None   # not populated yet
```

So doing `minkowski_tensors(tri._verts, tri._faces)` right after parsing passes `None` for both arguments. The correct workaround requires calling a private method first:

```python
tri._consolidate_lists()
result = minkowski_tensors(tri._verts, tri._faces, labels=tri._labels)
```

This is awkward for two reasons: it relies on a private method (`_consolidate_lists`) and private attributes (`_verts`, `_faces`, `_labels`), neither of which are part of the public API.

## Remaining inefficiency

Even with the workaround above, `minkowski_tensors` internally reconstructs a `Triangulation` from the extracted arrays at line 348:

```python
surface = Triangulation.from_arrays(verts, faces, labels=face_labels)
```

So a `Triangulation` is being torn down to arrays only to be rebuilt — wasting the consolidation work already done.

## Proposed fix

Modify `minkowski_tensors` to detect a `Triangulation` input, consolidate it if needed, extract the arrays for the existing preprocessing logic, and then **reuse the object directly as `surface`** — skipping the `from_arrays` rebuild:

```python
def minkowski_tensors(verts, faces=None, labels=None, ...):
    # Short-circuit for Triangulation input
    if isinstance(verts, Triangulation):
        tri = verts
        tri._consolidate_lists()
        if faces is None:
            faces = tri._faces
        if labels is None and tri._labels is not None:
            raw_labels = tri._labels
            if not np.all(raw_labels == LABEL_UNASSIGNED):
                labels = raw_labels
        verts = tri._verts
        # ... then let existing logic proceed, but pass tri directly as surface
    ...
    surface = tri if isinstance(...) else Triangulation.from_arrays(verts, faces, labels=face_labels)
```

This keeps the existing array-based API fully intact, requires no changes for current callers, and avoids the redundant rebuild when a `Triangulation` is passed.

### LINKED ISSUE #94: Wire check_surface() into the API: warn (or NaN) for open and non-manifold meshes to match C++ karambola behavior
## Background

`check_surface()` is fully implemented in `surface.py` — a complete Python equivalent of `check_surface.cpp` — but it is only called from the CLI (`cli.py`). API users calling `minkowski_tensors()` directly receive no warning about open surfaces or non-manifold meshes, and volume-dependent quantities (`w000`, `w020`) are computed silently even when they are geometrically undefined.

The test coverage added in PR #93 (issue #78) documents this gap. This issue tracks the fix.

---

### Open mesh handling

C++ karambola assigns each surface component a status (`closed`, `shared`, or `open`) and enforces a per-quantity `allowed_to_calc` threshold:

| Quantity | C++ karambola (open mesh) | pykarambola (current) |
|---|---|---|
| `w000`, `w010`, `w020` | **NaN / `cant_be_calculated`** | Computed silently — no warning |
| `w100`, `w110`, `w120`, `w102` | Computed normally | ✓ Computed normally |
| `w200`, `w300`, `w210`, `w310`, `w220`, `w320`, `w202` | NaN unless `--force` | Computed silently |
| Surface status in output | `open` / `shared` / `closed` keyword on every result line | Not surfaced via API |

### Degenerate triangle handling

| Situation | C++ karambola | pykarambola (current) |
|---|---|---|
| Small area ratio (`< 1e-12`) | Warning printed, continues | ✓ `UserWarning` raised, continues |
| NaN from `acos` (zero-length edge) | **Hard crash** (`std::exit(-1)`) | ✓ Stays finite — guarded safely |
| Zero angle-sum vertices | Not explicitly guarded | ✓ Guarded with warning and safe fallback |

### Non-manifold mesh handling

| C++ karambola | pykarambola (current) |
|---|---|
| `std::runtime_error`: "your polyfile is damaged" — aborts | No crash; computes silently |

---

## Proposed fix

Wire `check_surface()` into `minkowski_tensors()` (and `minkowski_tensors_from_label_image()` where applicable):

1. After `Triangulation` is built, call `check_surface()` (or an equivalent surface-status check) per label.
2. If a component is **open**: emit a `UserWarning` and either set `w000`/`w020` to `NaN` or document that values are unreliable. A `force=True` kwarg could preserve current behavior for users who need it.
3. If a component is **non-manifold**: emit a `UserWarning`. Do **not** abort (pykarambola's more permissive behavior here is an intentional improvement over C++).

The degenerate-triangle and zero-angle-sum guards are already correctly placed; no change needed there.

---

## Scope notes

- Non-manifold: keep the more permissive behavior (warn, don't abort) — matching C++'s hard crash is not desirable.
- The existing tests in `test_mesh_quality.py` (PR #93) already cover these paths. Once the fix lands, `TestOpenSurface` and `TestNonManifoldMesh` should be updated to assert the new warning behavior.

### LINKED ISSUE #99: minkowski_tensors_from_label_image: add pad parameter to prevent open surfaces at image boundary
## Problem

`minkowski_tensors_from_label_image` passes the label image directly to `marching_cubes` with no padding. If any labelled object touches the boundary of the array, marching cubes produces an **open (clipped) surface** at that face, which yields incorrect Minkowski tensor values for that object.

This is silently related to #94 — the open surface that `check_surface()` would detect is often caused by exactly this: the object was not padded before meshing.

## Proposed fix

Add a `pad` parameter (default `True`) that applies a 1-voxel zero-padding on all six faces before calling `marching_cubes`, then subtracts the padding offset from the resulting vertex coordinates so the output is in the original image coordinate system:

```python
def minkowski_tensors_from_label_image(
    label_image, level=None, spacing=(1.0, 1.0, 1.0),
    center='centroid_mesh', center_per_label=True,
    compute='standard', compute_eigensystems=True,
    return_count=False, autolabel=False,
    pad=True,   # <-- new parameter
):
```

Inside the function, right after `label_image = np.asarray(label_image)`:

```python
if pad:
    label_image = np.pad(label_image, pad_width=1, mode='constant', constant_values=0)
    _pad_offset = np.array(spacing)   # 1-voxel shift in each axis
else:
    _pad_offset = np.zeros(3)
```

Then subtract the offset from all vertex arrays returned by `marching_cubes`:

```python
verts, faces, _, _ = marching_cubes(...)
verts = verts - _pad_offset
```

This needs to be applied at every `marching_cubes` call site in the function (autolabel path, global centroid cache path, and per-label path).

### Why `pad=True` as default?

A 1-voxel zero pad is the minimum needed to guarantee a closed surface for any object, costs negligible memory, and matches the expectation of most users. Users who have already padded their image manually, or who intentionally want open surfaces, can pass `pad=False`.

yajushikhurana:
## Summary

- **#90** — `minkowski_tensors` now accepts a `Triangulation` object directly as its first argument (`faces` may be omitted). Handles both append-mode Triangulations from parsers and fully-built ones from `from_arrays`. All four parsers (`parse_obj_file`, `parse_off_file`, `parse_poly_file`, `parse_glb_file`) return compatible objects.
- **#94** — Surface quality checks are wired into `minkowski_tensors` via a new `_get_open_and_nonmanifold` helper: open surfaces emit a `UserWarning` and have `w000`/`w020` set to `NaN`; non-manifold meshes emit a `UserWarning` without aborting (intentionally more permissive than the C++ CLI).
- **#99** — `minkowski_tensors_from_label_image` gains a `pad=True` parameter that applies a 1-voxel zero-padding before `marching_cubes`, ensuring objects touching array boundaries produce closed surfaces; all vertex outputs and centroid_voxel computations are corrected for the offset.
- Version bumped `0.2.0 → 0.3.0`.

## Test plan

- [x] `TestOpenSurface`: asserts `UserWarning` with "open surface" in message; asserts `w000` is NaN; asserts `w020` is all-NaN; `w100` still finite/positive
- [x] `TestNonManifoldMesh`: asserts `UserWarning` with "non-manifold" in message using new `_bowtie_non_manifold_mesh` fixture (two closed fans sharing vertex 0); still no crash
- [x] `TestTriangulationInput` (5 new tests): `from_arrays` object accepted; results match raw array input; `parse_off_file` output accepted; `parse_obj_file` output accepted; `TypeError` raised when `faces=None` and `verts` is not a Triangulation
- [x] `TestBaselineClosed::test_no_warnings` still passes — no spurious warnings for clean closed meshes
- [x] All 148 existing tests pass (46 skipped due to optional deps)

Closes #90, closes #94, closes #99


keisuke-ishihara (changes requested):
## Review — PR #108 (Triangulation input, surface checks, label image padding)

**Overall:** No critical bugs; backward-compatible; well-tested for #90 and #94. Requesting changes for two gaps in test coverage and one dead-code issue before merge.

---

### Feature #90 — Triangulation input ✓

Polymorphic dispatch and label-extraction logic are correct. The case where a user passes both a labelled `Triangulation` *and* an explicit `labels=` arg is handled correctly but should be documented more explicitly in the docstring.

**Gap:** The CHANGELOG claims `parse_poly_file` / `parse_glb_file` outputs are supported, but no test exercises either parser with `minkowski_tensors(tri)`. Please add smoke tests matching the existing `parse_off_file` / `parse_obj_file` ones.

---

### Feature #94 — Surface quality checks ✓ (mostly)

Core logic is correct: open-edge detection via `NEIGHBOUR_UNASSIGNED` and the `sum_tris != len(tris)` fan-traversal check are both sound and tested with the bowtie fixture.

**Issue 1 — dead code in `_get_open_and_nonmanifold`:**

The condition

```python
if old_t == nxt:
    is_non_manifold = True
    break
```

is never triggered by any test and is unreachable in well-formed meshes (if it fired, it would mean we stepped backward in the fan, not forward). Either add a concrete test case that exercises this branch, or remove it to keep the traversal logic legible.

**Issue 2 — mixed error cases:**

```python
if vid == -1 or surface.ith_neighbour_of_triangle(cur_t, vid) == NEIGHBOUR_UNASSIGNED:
    neigh_un = True
    break
```

`vid == -1` signals a data-integrity error (vertex not found in a triangle that was supposed to contain it), while `== NEIGHBOUR_UNASSIGNED` signals a normal open edge. These are different conditions and should be handled separately — at minimum with a comment, ideally with an `assert vid != -1` guard before the open-edge check.

---

### Feature #99 — Padding ✓ (logic correct, tests missing)

The coordinate offset arithmetic (`_pad_offset = np.array(spacing)`) is correct for 1-voxel symmetric padding, including anisotropic spacing. However, there are **no dedicated tests** for the `pad` parameter. Please add at least:

```python
def test_pad_closes_boundary_touching_object(self):
    vol = np.zeros((10, 10, 10), dtype=int)
    vol[0:5, 0:5, 0:5] = 1  # touches three boundary faces
    with warnings.catch_warnings(record=True) as w:
        warnings.simplefilter("always")
        minkowski_tensors_from_label_image(vol, pad=True)
    assert not any("open surface" in str(x.message) for x in w)

def test_no_pad_boundary_touching_warns(self):
    vol = np.zeros((10, 10, 10), dtype=int)
    vol[0:5, 0:5, 0:5] = 1
    with pytest.warns(UserWarning, match="open surface"):
        minkowski_tensors_from_label_image(vol, pad=False)
```

---

### Summary of requested changes

| Priority | Item |
|----------|------|
| **Required** | Add `pad=True/False` tests for boundary-touching objects (#99) |
| **Required** | Remove or add test coverage for the dead `old_t == nxt` non-manifold branch (#94) |
| **Required** | Add `parse_poly_file` / `parse_glb_file` smoke tests for Triangulation input (#90) |
| Nice-to-have | Separate the `vid == -1` guard from the open-edge check with a comment or assert |
| Nice-to-have | Annotate `verts` param type as `Union[np.ndarray, Triangulation]` |

yajushikhurana:
All requested changes have been addressed:

**Required**
- Added `TestPadParameter` with `test_pad_closes_boundary_touching_object` and `test_no_pad_boundary_touching_warns` (#99)
- Removed dead `old_t == nxt` branch from `_get_open_and_nonmanifold`; also cleaned up the now-unused `old_t` variable (#94)
- Added `parse_poly_file` and `parse_glb_file` smoke tests to `TestTriangulationInput` (glb skipped when trimesh not installed) (#90)

**Nice-to-have**
- Separated `vid == -1` data-integrity guard from the open-edge check with an `assert` and explanatory comment
- Annotated `verts` parameter as `Union[np.ndarray, Triangulation]`

keisuke-ishihara (approved):

PR #113, opened 2026-04-29
Add example notebooks: NumPy array API and label-image API (closes #16, #17)

### LINKED ISSUE #16: Example notebook: NumPy array API walkthrough
A self-contained notebook demonstrating the `minkowski_functionals()` API from NumPy arrays.

## Content
- Load or generate a simple mesh (e.g. a sphere, a cube)
- Call `pk.minkowski_functionals(verts, faces)` and show the returned dict
- Explain each output key (w000 = volume, w100 = surface area, w020 = rank-2 tensor, etc.)
- Show the `compute=` argument for selecting specific quantities
- Show multi-label usage via the `labels=` argument

**Blocks:** #29

### LINKED ISSUE #17: Example notebook: label-image API — single-label vs multi-label on same 3D binary image
Demonstrate the label-image API using the same 3D binary image under two labeling schemes.

## Case (i) — single label
- Set all foreground voxels to label 1
- Call `minkowski_functionals_from_label_image()`
- Returns one result describing the whole structure as a single object

## Case (ii) — multiple labels
- Assign each connected component a unique label via `scipy.ndimage.label`
- Same function call, same underlying binary image
- Returns per-object Minkowski tensors, revealing individual shape differences

## Dataset
Use a real or representative synthetic 3D bioimage with multiple distinct objects.

## Output
Notebook is the basis for figure F6 and writing section W12.

**Blocks:** F6, W12

yajushikhurana:
## Summary

- **`examples/pykarambola_demo.ipynb`** — extends the existing demo to close #16: adds explicit `labels=` array usage, `labels='auto'` + `return_count`, and a new §10 covering `_beta`/`_trace`/`_trace_ratio` derived scalars and `compute_eigensystems=False`
- **`examples/label_image_api.ipynb`** — new dedicated notebook closing #17: demonstrates single-label vs multi-label on a real AllenCell hiPSC segmentation (Anaphase/Telophase/Cytokinesis Early)
- **`pyproject.toml`** — adds `tifffile` to the `notebooks` extra
- **`dataset/multilabel_test/f5d35d71_segmentation.ome.tif`** — 2-channel OME-TIFF segmentation used by `label_image_api.ipynb` (nuclei + cell body, AllenCell open data)

### label_image_api.ipynb cases

| Case | Label assignment | What is measured |
|------|-----------------|-----------------|
| 1 | Cell body → label 1 | Whole-cell morphometry |
| 2 | Nuclei → label 1, cell body → label 2 | Per-compartment morphometry |
| 3 | `scipy.ndimage.label` on nuclei channel | Per-nucleus anisotropy — β values compared across the 3 detected components |

Visualisation uses matplotlib mid-Z cross-sections and an anisotropy bar chart; no display-server or heavy 3D rendering dependencies required.

## Test plan
- [ ] Run `examples/pykarambola_demo.ipynb` end-to-end and verify all new cells produce expected output
- [ ] Run `examples/label_image_api.ipynb` end-to-end; confirm 3 components detected in Case 3
- [ ] Confirm `pip install "pykarambola[notebooks]"` now pulls in `tifffile`


keisuke-ishihara (commented):
## Review

Overall: solid notebooks that cover the API well and are clearly executable. A few things to address before merging.

---

### `pyproject.toml`
Clean — adding `tifffile` to the `notebooks` extra is correct.

---

### `examples/pykarambola_demo.ipynb`

**§5 — negative volume output is unexplained.**
Cell `cell-14` outputs `Volume (w000): -24.0000` for both labels, with three UserWarning messages about non-manifold mesh and inverted normals. There is no surrounding markdown explaining that this is expected behavior from the `two_boxes_sharing_vertex_with_labels.poly` test fixture (shared vertex = non-manifold). A reader will be confused. Consider either:
- Adding a markdown note explaining this is an intentionally degenerate fixture used to show warning behavior, or
- Replacing the fixture with the clean programmatically-built two-box example from §5b (which gives correct positive volumes)

**§10 — `_trace_ratio` rows print `nan` without explanation.**
This is technically correct (`w102` and `w202` have no paired scalar), but no comment or table note explains why. Worth a one-liner: `# w102, w202 have no paired scalar → trace_ratio undefined`.

---

### `examples/label_image_api.ipynb`

**Case 3 — commented-out debug code should be removed** (cell `4ba23e94`):
```python
# axes[0].imshow(label_case3[100], cmap=cmap3, vmin=0, vmax=n_nuclei, interpolation='nearest')
# axes[0].set_title(f'Per-nuclear object labels  |   Z-slice {100}')
```
These lines belong to a prior exploratory pass and clutter the final notebook.

**Case 3 — `mid_z=48` only shows 2 of 3 nuclear objects.**
Objects 2 and 3 span Z=40–72 and Z=78–110 respectively — at `mid_z=48`, Object 3 is completely absent from the label-map image. The bar chart correctly shows all 3 objects, but the visualization is misleading. Either:
- Change `mid_z` to a slice that intersects all three (Z ≈ 55–70 covers objects 1 and 2; no single slice covers all three given their non-overlapping Z ranges — a note explaining this would suffice), or
- Add a brief note that slice 48 shows only objects 1 and 2 because object 3 is axially separated

**`plt.cm.get_cmap` is deprecated since matplotlib 3.7** (cells `e04a0fbf` and `4ba23e94`):
```python
cmap2 = plt.cm.get_cmap('Set1', 3)          # DeprecationWarning in mpl ≥ 3.7
cmap3 = plt.cm.get_cmap('tab10', n_nuclei + 1)
```
Replace with:
```python
import matplotlib
cmap2 = matplotlib.colormaps['Set1'].resampled(3)
cmap3 = matplotlib.colormaps['tab10'].resampled(n_nuclei + 1)
```

---

### `dataset/multilabel_test/f5d35d71_segmentation.ome.tif`

At 205 KB this is small enough to commit directly. Fine as-is.

---

### Summary

| # | Severity | Item |
|---|----------|------|
| 1 | Medium | demo §5: unexplained negative volume + warnings |
| 2 | Minor | label_image `mid_z=48` hides Object 3 from label map |
| 3 | Minor | label_image: commented-out debug lines in Case 3 cell |
| 4 | Minor | label_image: `plt.cm.get_cmap` deprecated call ×2 |
| 5 | Nit | demo §10: no explanation for `nan` trace_ratio rows |

keisuke-ishihara:
## Comments on `pykarambola_demo.ipynb`

### 1. Title and framing: Minkowski *tensors*, not *functionals*

The intro section title says **"What are Minkowski functionals?"** — it should be **"What are Minkowski tensors?"** Minkowski *functionals* specifically refers to the rank-0 scalar members (w000, w100, w200, w300). The tensors are the general family; functionals are the special case. The section should make this distinction explicit.

### 2. Example 1 — box surface area: why 17.333, not 52?

The true surface area of the 2×3×4 box is `2·(ab + bc + ca) = 52`. But `w100 = 17.333 = 52/3`. The formula in the comment `2/3*(a*b + b*c + a*c)` is algebraically equivalent to `A/3` — it just obscures this. The comment should say `A/3, where A = 52`.

### 3. Why does karambola divide by 3? — the Steiner formula

The normalization comes directly from the **Steiner formula** for parallel bodies. The volume of the ε-dilation K⊕εB of a body K expands as a polynomial in ε:

$$V(K \oplus \varepsilon B) = \sum_{\nu=0}^{3} \binom{3}{\nu} W_\nu(K)\, \varepsilon^\nu = W_0 + 3W_1\varepsilon + 3W_2\varepsilon^2 + W_3\varepsilon^3$$

In geometric terms this is simply `V + A·ε + M·ε² + (4π/3)·ε³`. Matching coefficients:

| ν | Steiner term | Binomial | Minkowski functional | Geometric meaning |
|---|---|---|---|---|
| 0 | W₀ | 1 | `w000 = V` | volume |
| 1 | 3·W₁ = A | 3 | `w100 = A/3` | surface area / 3 |
| 2 | 3·W₂ = M | 3 | `w200 = M/3` | integrated mean curvature / 3 |
| 3 | W₃ | 1 | `w300 = 4π/3` for convex body | see below |

The factors of **3 = binom(3,1) = binom(3,2)** are not an arbitrary choice — they are the binomial coefficients that appear in the Steiner formula definition of Minkowski functionals in d=3. This is the standard **Hadwiger/Blaschke normalization** from integral geometry (see Schröder-Turk et al. 2013, NJP).

### 4. Euler characteristic — `w300` convention

For a topological sphere (box surface), V − E + F = 8 − 12 + 6 = **χ = 2**. But `w300 = 4π/3 ≈ 4.1888 ≠ 2`. The relationship is:

```
w300 = (2π/3) · χ
```

So for χ = 2: `w300 = 4π/3 ✓`. The notebook currently says `expected 4pi/3` without explaining that this encodes χ = 2 with a prefactor of 2π/3. This should be made explicit.

### 5. Section 2 — naming

- "Scalar functionals" → "Rank-0 tensors (scalars)"
- "Vector functionals" → "Rank-1 tensors (vectors)"

Consistent with the broader correction above.

### 6. Example 6 — sphere surface area discrepancy

Smooth sphere with r = 20: `4π·20² = 5026.5`, so the **expected** `w100 = A/3 ≈ 1675.5`. The notebook prints `w100 = 1817.6` with no comparison and no mention of the `A/3` convention. Two things should be added: (a) an analytical `w100` comparison parallel to the volume comparison, and (b) a note that the ~8.5% area overestimate (vs. ~0.5% for volume) is expected — marching cubes surface area on a voxel grid converges much more slowly than volume due to staircase artefacts.

keisuke-ishihara:
## Normalization convention applies to all tensor ranks

The `1/binom(3,ν)` factor from the Steiner formula is not specific to the scalar functionals — it applies uniformly across the entire Minkowski tensor family. The general tensor W^{r,s}_ν (r position weights, s normal weights, order ν) carries the same normalization as its scalar counterpart W^{0,0}_ν:

| ν | binom(3,ν) | Factor | Affected quantities |
|---|---|---|---|
| 0 | 1 | ×1 | w000, w010, w020 |
| 1 | 3 | ×1/3 | w100, w110, w120, w102 |
| 2 | 3 | ×1/3 | w200, w210, w220, w202 |
| 3 | 1 | ×1 (+ 2π/3 prefactor) | w300, w310, w320 |

This can be verified directly from the box output. For **w102** (ν=1, normal-weighted rank-2): the six faces contribute `∫ n⊗n dA = diag(24, 16, 12)` (e.g. the ±x faces each have area 12, normal ±x̂, contributing 2·12 = 24 to the xx component). Dividing by 3 gives `diag(8, 5.33, 4)`, matching the notebook eigenvalues `[4.000, 5.333, 8.000]` exactly.

The practical implication for the notebook: whenever a tensor's trace or eigenvalues are compared to a geometric quantity, the factor must be accounted for. For example:
- `Tr(w120) = (1/3) ∫_∂K |x|² dA`, not the raw moment
- `Tr(w102) = (1/3) · surface area = w100` (since `∫ n⊗n dA` has trace equal to the surface area, by `|n|²=1`)

keisuke-ishihara:
@yajushikhurana I always get the normalization convention confused. 

Let's fact check the basis of the normalization, and incorporate this info in the notebook, so we don't have to go back to Schroeder-Turk all the time.

yajushikhurana:
Thanks for the thorough review! Here's how we're addressing each point:

**Accepted — fixed in this update:**

**1. Negative volume + warnings (demo §5):** Added inline comments to the `.poly` fixture cell explaining that inverted face winding produces `w000 < 0` (true volume = `abs(w000)`) and that the shared vertex is intentional non-manifold geometry that triggers the warnings. The print statement now shows `true vol = 24.0000` alongside the raw negative value.

**4. `plt.cm.get_cmap` deprecation ×2:** Both calls replaced with `plt.colormaps['Set1'].resampled(3)` / `plt.colormaps['tab10'].resampled(n_nuclei + 1)`.

**5. NaN trace_ratio rows (demo §10):** Added a callout note to the §10 markdown cell: `_trace_ratio` is only defined for position-weighted wX20 tensors (w020, w120, w220, w320); the normal-weighted w102 and w202 have no natural paired Minkowski scalar, so their `_trace_ratio` is always `nan`.

---

**Points 2 and 3 — resolved together via two-slice visualization:**

**2. `mid_z=48` hides Object 3:** The three nuclear objects do not share a common Z-slice where all three are visible simultaneously — Objects 2 and 3 occupy different axial regions (Z 40–72 vs 78–110, as shown in the table in the preceding markdown cell). Rather than picking a single mid-Z that hides one object, we now show **two Z-slice panels side by side**: Z=55 (Objects 1 & 2 visible) and Z=94 (Objects 1 & 3 visible), followed by the anisotropy bar chart — a 1×3 subplot layout.

**3. Commented-out lines in Case 3 cell:** Those lines (`# axes[0].imshow(label_case3[100], ...)`) were intentional inline documentation of the data structure (specifically, why you can't see all three objects in one frame). With the two-slice approach above, those comments are no longer needed and have been replaced by the explicit two-panel display.

yajushikhurana:
Follow-up changes pushed:

- **pykarambola_demo.ipynb**: expanded Steiner normalization cell to cover all tensor ranks with a full table; added χ=2−2g genus interpretation to the Gauss–Bonnet callout; linked to new reference doc
- **docs/minkowski_tensors.md** (new): full reference for all 16 quantities — LaTeX notation, analytical integrals, discrete computational formulas from pykarambola source (including explicit w010 triangle polynomial), and physical interpretations
- **README.md**: added link to `docs/minkowski_tensors.md` under Computed quantities

keisuke-ishihara (commented):
Good overall structure — the two notebooks cover the API well and the Steiner normalization section is a great addition. Three things need fixing before merge, plus a few minor items.

---

## 1. `docs/minkowski_tensors.md` — factual error in the Steiner table description

> "The $1/3$ and $1/6$ prefactors are the binomial normalizations $1/\binom{3,\nu}$ from the Steiner formula"

This is wrong for the edge weight. $\binom{3}{2} = 3$, so $1/\binom{3}{2} = 1/3 \neq 1/6$. The edge prefactor $1/6$ combines two separate factors:

$$\frac{1}{6} = \underbrace{\frac{1}{\binom{3}{2}}}_{\text{Steiner}} \times \underbrace{\frac{1}{2}}_{\text{curvature: }H_e = \tfrac{1}{2}\alpha_e \ell_e}$$

The face weight $1/3 = 1/\binom{3}{1}$ is purely Steiner ✓, but the edge weight is not. The sentence should read something like:

> "The face weight $1/3 = 1/\binom{3}{1}$ is the Steiner normalization. The edge weight $1/6 = (1/\binom{3}{2}) \times (1/2)$ combines the Steiner factor with the $1/2$ from the discrete curvature formula $H_e = \tfrac{1}{2}\alpha_e\ell_e$."

Similarly, the vertex weight $1/3$ in $w_{300} = \tfrac{1}{3}\sum_v \delta_v$ is not $1/\binom{3}{3} = 1$; it comes from the definition $W_3 = \tfrac{1}{3}\int K_G\,dA$ (Hadwiger convention).

---

## 2. `label_image_api.ipynb` — `Surface area (w100)` is mislabeled throughout

In Cases 1, 2, and 3, code like:

```python
print(f'  Surface area (w100):       {r["w100"] * voxel_size_um**2:.1f} µm²')
```

prints $w_{100} \times \text{voxel\_size}^2$, which equals $A/3$ in µm², **not** the surface area $A$. A user seeing "Surface area: 499.4 µm²" would assume the cell's surface area is 499.4 µm² when the true value is ~1498 µm².

Fix: either

```python
print(f'  Surface area (3·w100):     {r["w100"] * voxel_size_um**2 * 3:.1f} µm²')
```

or label it as `w100 (= A/3)` consistently. The demo notebook is explicit about this (`w100 = A/3`) — the label-image notebook should be too.

---

## 3. `label_image_api.ipynb` — `Euler characteristic` prints `w300`, not χ

```python
print(f'  Euler characteristic:      {r["w300"]:.4f}')
```

`w300 = (2\pi/3)\chi`, so for a topological sphere the output is 4.1888, which is not χ = 2. Should be either:

```python
print(f'  w300 (= 2πχ/3):           {r["w300"]:.4f}')
# or
import math
print(f'  Euler characteristic χ:    {round(3 * r["w300"] / (2*math.pi))}')
```

---

## Minor

**Demo notebook Cell 28**: `LABEL_UNASSIGNED` is imported but never referenced in the cell. Remove it.

**Case 2 naming**: Label 2 is the cytoplasm (cell body voxels where nuclei took priority are excluded), not the full cell body. The comment `# cell body → label 2` and the print label `'cell body'` are slightly misleading. Labelling it `'cell body (excl. nuclei)'` or `'cytoplasm'` — and noting that volumes satisfy `V_case2_label2 ≈ V_case1 - V_nuclei` — would help readers understand what is being measured.

**README**: Extra blank line introduced before the quantities table (line ~206 in the diff).

---

Everything else looks good: the two-slice visualization for Case 3 is a clean solution, the `center=None` explanation in Case 3b is clear, the additivity demo is mathematically sound, and the `docs/minkowski_tensors.md` reference is a genuinely useful addition once the Steiner sentence is corrected.

keisuke-ishihara:
The above was the Claude review.

Here is what I noticed:
1. I caught the inconsistency between Euler Char and w300. Please fix that one.
2. In the label_image_api.ipynb, as a cell biologist, I'm a bit bothered by calling the individual objects "Case 3 — Per-nucleus: individual nucleus morphometry". If the images are mostly interphase, that's fine, but during mitosis, there is no nucleus, or multiple of them. I know we've been calling it the 'nucleus' channel, but maybe we should call it DNA channel. For this notebook, maybe just calling it "per object" individual object morphometry" etc. might be more approriate. Let's just stop calling this data set nucleus.

yajushikhurana:
> 2. In the label_image_api.ipynb, as a cell biologist, I'm a bit bothered by calling the individual objects "Case 3 — Per-nucleus: individual nucleus morphometry". If the images are mostly interphase, that's fine, but during mitosis, there is no nucleus, or multiple of them. I know we've been calling it the 'nucleus' channel, but maybe we should call it DNA channel. For this notebook, maybe just calling it "per object" individual object morphometry" etc. might be more approriate. Let's just stop calling this data set nucleus.

Haha, I have manually updated this many times only for Claude to revert it back, I have been using nuclear objects instead of nucleus, but anytime I ask it to make a new change it sneaks the per nucleus in oof! I guess maybe modifying the claude.md file to not over write my changes or something, or just asking Claude to make these changes and not do them manually would be the way to go!

yajushikhurana:
@keisuke-ishihara Pushed another round of fixes addressing the latest review:

- **Case 2** now uses two separate calls (one per channel) so the cell body is measured whole, not as cytoplasm
- **Channel 0** renamed from "nuclei" → "DNA" throughout the label-image notebook
- **Case 3** renamed to "Per-nuclear object" / "nuclear object" throughout
- **Surface area** prints now labeled `3·w100 = A` to make the A/3 convention explicit
- **Euler characteristic** now prints χ alongside the raw w300 value (`χ = round(3·w300 / 2π)`)
- **`docs/minkowski_tensors.md`**: corrected the edge weight explanation — 1/6 = (1/binom(3,2)) × (1/2 from discrete curvature H_e), not purely a Steiner normalization; vertex weight note also fixed
- **Demo notebook Cell 28**: removed unused `LABEL_UNASSIGNED` import
- **README**: removed extra blank line before the computed quantities table

keisuke-ishihara:
For `label_image_api.ipynb`,
image

I think there is a mismatch between the code and the printed output. Maybe you have to run the notebook before commiting?

Case 3 renamed to "Per-nuclear object" / "nuclear object" throughout
-> Let's avoid saying "nuclei" or "nuclear" for the entire notebook? Just say DNA or simply "objects".

keisuke-ishihara:
also, from a pedagogical point of view, can you introduce an example with spheres of two different raddii in `pykarambola_demo.ipynb` before the box? The Steiner normalization is supposed to facilitate shape analysis.

I think if you take the ratio of w000/w100 this should be constant for different sized spheres with the normalization (good to demonstrate this), while the the typical $`\frac{4\pi/3 r^3}{4 \pi r^2}=\dfrac{r}{3}`$ is scale-dependent.

now that I write it like this, I understand why the 1/3 takes care of the cofactor, but w000/w100 = r still, I think you can educate me on this.

yajushikhurana:
Pushed fixes for both comments:

**`label_image_api.ipynb`** — all "nuclei"/"nuclear" language removed: variable renamed `nuclei_mask` → `dna_mask`, markdown, print labels, plot titles, and summary table now use "DNA objects" or "objects" throughout. Notebook re-executed to fix the output mismatch.

**`pykarambola_demo.ipynb`** — added a sphere comparison section before the box. On your question: `w000/w100 = r` is indeed *not* constant — it gives a scale-dependent characteristic length. What *is* constant across sphere sizes is the dimensionless isoperimetric ratio `w000²/w100³ = 3/(4π) ≈ 0.2387`. The new cells show this numerically for r=5 and r=20, and explain the distinction in a markdown cell with the formula. The Steiner normalization doesn't produce scale invariance on its own — it puts all Wν on a consistent dimensional footing so you *can* form meaningful dimensionless ratios.

Branch also rebased onto main to resolve the `pyproject.toml` conflict from #120 (kept `stl = ["numpy-stl"]` from main, kept `tifffile` in `notebooks` from this branch).

yajushikhurana:
@keisuke-ishihara The demo notebook has been reordered to keep the normalization explanation in the beginning.

keisuke-ishihara:
## Review audit

All items from both review rounds have been addressed. Summary below.

### Review 1 (2026-04-30) — 5/5 items addressed

| # | Item | Status |
|---|------|--------|
| 1 | Demo §5: explain negative volume + non-manifold warning | Done — comment added explaining inverted winding / non-manifold fixture |
| 2 | Case 3: `mid_z=48` hides Object 3 | Done — replaced with two-slice layout (Z=55 shows obj 1&2; Z=94 shows obj 1&3) |
| 3 | Case 3: remove commented-out debug lines | Done — removed |
| 4 | `plt.cm.get_cmap` deprecated ×2 | Done — replaced with `plt.colormaps[...].resampled(n)` |
| 5 | Demo §10: explain `nan` trace_ratio rows | Done — callout note added explaining w102/w202 have no paired scalar |

### Review 2 (2026-05-03) — 6/6 items addressed

| # | Item | Status |
|---|------|--------|
| 1 | `docs/minkowski_tensors.md`: edge weight 1/6 explanation wrong | Done — corrected to `1/6 = (1/binom(3,2)) × (1/2 from H_e)` |
| 2 | Surface area mislabeled as `w100` instead of `3·w100` | Done — now prints `Surface area (3·w100 = A)` with ×3 applied |
| 3 | Euler char prints `w300` instead of χ | Done — now prints `χ = round(3·w300 / 2π)` alongside w300 |
| 4 | Cell 28: unused `LABEL_UNASSIGNED` import | Done — removed |
| 5 | Case 2: "cell body" vs. cytoplasm distinction | Done — uses separate calls per channel so cell body is measured whole; prose notes the relationship to Case 1 |
| 6 | README: extra blank line before quantities table | Done — removed |

Ready to approve.

keisuke-ishihara:
This looks great, thank you! will merge

keisuke-ishihara:
oh i noticed one things - minkowski_tensors.md
i think the latex is not rendering correctly on github

yajushikhurana:
@keisuke-ishihara the markdown file should be corrected now. I was using markdown preview enhanced plugin for VS code that renders latex better in markdown files but also bypasses those errors. 

In GitHub Markdown table cells, _ at word boundaries like }_{ and }_i gets consumed by the Markdown italic parser before KaTeX processes the math. The _ vanishes, leaving bare subscript targets like {ij} or just i floating outside the math. The fix: escape those underscores as \_ specifically inside table cells. Outside tables (in the $$ blocks) they're fine. 

The w120 and w202 $$ blocks have lines starting with - which GitHub parses as list items. Put each on one line corrects the issues in those sections

keisuke-ishihara (approved):
Thanks for the fix!

PR #115, opened 2026-05-04
Add end-to-end 3D segmentation → Minkowski tensor notebook (closes #84)

### LINKED ISSUE #84: Example notebook: 3D segmentation to Minkowski tensor analysis with mesh visualisation
## Goal

Demonstrate an end-to-end analysis pipeline starting from a raw 3D image through to colour-coded mesh visualisation driven by Minkowski tensor values.

## Notebook outline

1. **Load a 3D image** — use a synthetic or publicly available 3D fluorescence image.
2. **Segment objects** — apply Otsu thresholding (or connected-component labelling) to produce a label image.
3. **Extract meshes and compute Minkowski tensors** — call \`minkowski_tensors_from_label_image()\` to triangulate each labelled object and return its tensors in a single call.
4. **Colour-code by individual tensor values** — visualise meshes in napari or PyVista, mapping a scalar tensor quantity (e.g. \`w200\` mean curvature, anisotropy \`beta\`) to a per-object colour.
5. **PCA + clustering on the full tensor feature vector** — stack all per-object tensor outputs into a feature matrix, run PCA for dimensionality reduction, cluster (e.g. k-means), and colour-code meshes by cluster assignment.

## Why this matters

This notebook serves as a showcase for the paper and as a practical starting point for users who want to apply Minkowski tensor analysis to their own segmentation data.

## Related

- Builds on #17 (label-image API walkthrough notebook)
- Relevant to Results section (issue #33)

yajushikhurana:
## Summary

- Adds `examples/segmentation_to_tensors.ipynb`: a paper-showcase notebook taking `skimage.data.cells3d()` from raw confocal stack through to PCA + K-Means clustering on Minkowski tensor features
- Adds `scikit-learn` and `pooch` to the `notebooks` extra in `pyproject.toml`

## Notebook pipeline

1. Load `cells3d()` — 60-slice HeLa confocal stack, nuclei channel, physical voxel spacing 0.29 × 0.2645 × 0.2645 µm
2. Gaussian smoothing → Otsu threshold → fill holes → connected-component labelling → small-object filtering
3. `pk.minkowski_tensors_from_label_image(..., compute="all", spacing=SPACING)` — all scalars and rank-2 tensors in physical units
4. Colour-coded 3D mesh visualisation (`Poly3DCollection`) coloured by β anisotropy (w020_beta)
5. PCA on 8 Minkowski features + K-Means (k=2) clustering
6. Re-render meshes coloured by cluster assignment
7. Summary table of pipeline steps and tensor interpretations

## Test plan

- [ ] `conda run -n pykarambola-0.3.0 pip install -e ".[dev,accel,notebooks]"` installs cleanly
- [ ] `jupyter nbconvert --to notebook --execute examples/segmentation_to_tensors.ipynb` completes without error
- [ ] All 14 cells produce output (verified locally)

Closes #84


keisuke-ishihara (commented):
this looks pretty good! it really shows how you can go from images to Minkowski tensors very easily.

some comments:
1. segmentation - some nuclei are touching each other, cant we just use a standard watershed algorithm to resolve them? or are you intentionally keeping them stuck together to show the different in morphological features?
2. Computing Minkowski Tensors - in the printed table, i think we should display the volume and area with and without the Steiner normalization, just to remind people that the immediate output of our function is normalized
3. mesh visualization - i think you can set the scale of z so that they don't look this elongated, also try to use the xyz axis orientation so it's closer to the xy image
4. if you change the segmentation, the PCA plot will be different. The current issue with this is that the PCA plot does not make a biologically obvious conclusion. The goal of the notebook should be to convince biologists about the MinT features.

keisuke-ishihara (changes requested):

yajushikhurana:
Thanks for the detailed review! Here is a summary of what was addressed:

1. **Segmentation** — watershed alone was not sufficient to separate all touching nuclei cleanly. I switched to a hybrid watershed surface (−nuclei + 0.5 × membrane) which uses the membrane channel to guide boundaries between adjacent cells, combined with per-Z-slice `binary_fill_holes` per nucleus label to fix topological holes (negative Euler characteristics) that 3D fill-holes missed.

2. **Minkowski tensor table** — updated to display both the raw normalised output (w000, w100 with Steiner factor) and the derived physical quantities (volume in µm³, area in µm²) side by side, so the normalisation convention is explicit.

3. **Mesh visualisation** — Z axis is now scaled to physical aspect ratio using `set_box_aspect`, and axes are reordered to XYZ orientation to match standard microscopy images.

4. **PCA / clustering** — expanded the feature set to all 62 rotation- and translation-invariant Minkowski quantities and switched to Agglomerative Clustering. In addition to the requested changes, we also added a separate analysis section that removes nuclei touching the XY border, leaving only fully-captured interior nuclei. This allows clean cell-cycle-stage separation through PCA .

keisuke-ishihara (commented):
thanks @yajushikhurana, the final clustering result looks great!

just two small comments/Qs:
1. you can filter out the objects touching the border much earlier in the notebook
I think a good place to do it will be in cell 3 when you ``` # Remove small fragments; renumber directly to preserve watershed boundaries```. This will also make the notebook shorter and more to the point.
 
2. Did you have to try different combinations of features to get a nice clustering at the end? for example, how does the current result compare with
  - 4 scalars only
  - 4 scalars + betas
  - 4 scalars + eigenvalues

3. I would also avoid using "rotation-invariant" language here. It is true, but I don't want to make big deal out of it until we do a full characterization of rotation-invariant feature construction that #102 is addressing. In this notebook, just motivating the domain knowledge from Schroder-Turk et al. that beta, eigenvalues are commonly used in physics, is sufficient.

yajushikhurana:
I will get on to these changes regarding some of the questions you asked, 
> 2. Did you have to try different combinations of features to get a nice clustering at the end? for example, how does the current result compare with
> 
> * 4 scalars only
> * 4 scalars + betas
> * 4 scalars + eigenvalues
> 
Claude decided to use a very random subset initially, it was a partial set of the rotation invariant quantities and that did not perform very well, I then decided to use the current set. I can update and see how it does for all these three cases. 
> 3. I would also avoid using "rotation-invariant" language here. It is true, but I don't want to make big deal out of it until we do a full characterization of rotation-invariant feature construction that #102 is addressing. In this notebook, just motivating the domain knowledge from Schroder-Turk et al. that beta, eigenvalues are commonly used in physics, is sufficient.

Sure, I can avoid that for now. I decided to not use the vectors because we hadn't aligned things yet and wanted to have a description that doesn't necessarily go into the alignment of things yet, but I can find a smarter way to state that.

yajushikhurana:
@keisuke-ishihara  Addressing the latest round of feedback:

- XY-border nuclei are now removed during the segmentation step itself (no longer a separate notebook section), keeping the pipeline concise
- Added a 2×2 feature-subset comparison showing PCA + mesh plots for 4 scalars / +β / +eigenvalues / full descriptor set (k=3), so the reader can see how clustering improves with richer features
- "Rotation-invariant" is no longer used as a label or identifier for the feature group throughout the notebook; features are referred to by their explicit names (β, eigenvalues, trace, msm_ql, msm_wl). The term appears only briefly in the closing summary.

keisuke-ishihara (changes requested):
Almost there! One remaining item from the previous round:

**Cell 12 — "rotationally invariant" language**

The sentence:
> Crucially, all of these quantities are rotationally invariant, making them insensitive to the orientation of a given cell within the field of view.

still needs to be removed. As discussed, let's hold off on making rotation invariance a selling point until #102 fully characterises it. Motivating the features by their physics grounding (Schröder-Turk et al.) is sufficient here.

keisuke-ishihara:
One more request for the Minkowski tensor table (cell 6):

The current **χ (w300) [raw]** column header conflates the normalised tensor output with the topological Euler characteristic. Please split this into two columns:

1. **`w300 [raw]`** — the raw pykarambola output as-is
2. **`χ`** — the physical Euler characteristic, computed as `w300 × factor`, whose values should be close to integers (χ = 2 for a convex/sphere-like nucleus, 0 for a toroidal shape, etc.)

Reserve the χ notation strictly for the integer-valued Euler characteristic; using it for the normalised w300 is misleading to readers who know the topology.

yajushikhurana:
Addressing the latest round of feedback:

- **w300 / χ column split** — the Minkowski tensor table (cell 6) now has two separate columns: `w300 [raw]` for the Steiner-normalised pykarambola output and `χ [= w300×3/2π]` for the integer-valued Euler characteristic, keeping χ strictly for the topological quantity (χ ≈ 2 for convex/sphere-like nuclei)

- **Rotation-invariant language removed** — replaced with: *"The feature set is restricted to quantities whose geometric meaning is grounded in the intrinsic surface geometry of the cell (Schröder-Turk et al. 2013) rather than its placement within the image. Position-weighted tensors and vector quantities are excluded because their values change with the coordinate frame, which can confound morphological comparisons across cells acquired in different orientations."*

- **README** — added an Example notebooks section with a link table for both notebooks, and a 2×2 pipeline visual (raw confocal → segmented nuclei → PCA clustering → meshes coloured by cluster) showing the full-descriptor feature set (scalars + β + eigvals + trace + msm_ql/wl)

keisuke-ishihara (approved):
Thank you for the changes!
I like the snippet in the README!

PR #131, opened 2026-05-25
paper: fix benchmark methodology — align compute scope, expand to 121 features

yajushikhurana:
Closes #129

### Accuracy benchmark (`benchmark_adrenal3d.ipynb`)
- Aligned compute scope: re-ran karambola with MSM + w103/w104 to match pykarambola `compute='all'`
- Expanded from ~76 to **121 features**: w103 (27 components), w104 eigenvalues (6, via v0.5.0 API), msm_ql (9), msm_wl (9)
- Eigenvectors excluded from comparison: sign ambiguity and degenerate-eigenspace rotation make direct comparison ill-posed; eigenvalues fully characterise spectral geometry
- Replaced relative-error framework with `abs_diff_summary` (max and mean absolute difference per feature) — relative error broke for near-zero features regardless of threshold choice
- Primary accuracy figure: **2×4 scatter grid** (4 feature groups × 2 resolutions), red identity line — all 121 features from all 1,584 meshes fall on the line; follows Pachter et al. edgePython convention
- Bar chart: max absolute difference on log scale (absolute values, not normalised)
- Heatmaps: mean absolute difference on linear scale, split into 4 feature-group figures

### Runtime benchmark (`runtime_comparison.ipynb`)
- Re-ran karambola on icospheres **without MSM** to match pykarambola `compute='standard'` scope (closes #129)
- Corrected 28³ adrenal speedup: **14.6× → 2.8×** — the old 150 s figure included heavy per-mesh file I/O; 31 s is the corrected karambola wall time
- 64³ speedup unchanged at 1.6×

### Manuscript (`paper/manuscript.tex`)
- Feature count updated to 121 throughout
- Accuracy claim: "fall on the identity line" (supported by scatter figure)
- Eigenvector exclusion stated explicitly with justification
- All runtime numbers corrected

---

## Reviewer notes — please read

This PR took considerably more time and iterations than it should have. The methodology went through several pivots:
1. Relative error introduced → discovered it breaks for near-zero features
2. Switched to normalised absolute difference (max_abs_diff / max_val) → same problem with near-zero denominators
3. Finally settled on raw absolute differences on log scale + scatter plots as primary evidence

Claude made repeated errors during development: wrong feature counts in early drafts, a "machine epsilon" accuracy claim that did not hold for large-magnitude features (w220 reaches ~1.7×10⁻³ absolute difference), stale markdown cell headers not caught until explicitly flagged, and runtime numbers not updated until the human reviewer pointed out the discrepancy. **The human reviewer caught every one of these and pushed back each time.**

A strict review of:
- All numbers quoted in `manuscript.tex` (features counts, speedup figures, Cython time savings)
- All notebook cell headers and markdown text
- The `abs_diff_summary` logic and `compare_cols` completeness

…is strongly recommended before merging in case Claude shifted these changes again after being asked to make small changes to the notebook or manuscript. Claude was not reliable about consistency between the notebook and the manuscript without repeated prompting. Claude Code is useful for API edits etc, but can introduce errors and false explanations when actively interacting with Jupyter notebooks.


yajushikhurana:
Two candidate timing figure layouts have been committed in this second commit (`318c077`):

- **Option A** (`timing_comparison_option_a.pdf`): two-panel figure — left panel is absolute runtime on a log-log scale showing all four line series (pykarambola compute, pykarambola end-to-end, Cython compute, karambola); right panel is a speedup ratio (karambola / pykarambola) vs mesh size.
- **Option B** (`timing_comparison_option_b.pdf`): single two-panel figure (vertices and subdivision x-axes) showing only parse+compute end-to-end lines with annotated speedup factors at each point.

`timing_comparison.pdf` (the original single-panel log-log figure) is also still present.

Happy to rework the layout, change axis scaling, combine panels differently, or generate anything else the reviewer thinks would communicate the runtime story more clearly for the manuscript. Just leave a suggestion here or in the review.

keisuke-ishihara (commented):
Oh wow! This sounds like it was an epic battle w claude. 

I will review later this week. 

By the way, does this API change affect the pykarambola code and warrant a version bump?

@yajushikhurana

yajushikhurana:
> By the way, does this API change affect the pykarambola code and warrant a version bump?
> 
> @yajushikhurana

I think this can be a part of 0.4.0.

keisuke-ishihara:
0.4.0 is already released so staged for 0.5.0 then?

keisuke-ishihara:
re: versions
Ok, I think you did the current benchmarks using your pykarambola installation 0.3.0.
In the last two weeks, we had a version bump to 0.4.0 for the STL file parsers.
And now, you made an API change about w104 (which prompts a version bump to 0.5.0)

I think the cleanest solution is to isolate the 0.5.0 version bump as a separate PR for #130. (Remove #130 from the scope of this PR).
We let that merge into main and officially release 0.5.0 on pypi.
Then, you can do the current benchmark (rerun the pykarambola side if you need to) and then we can say we used pykarambola 0.5.0 in the manuscript latex file.

What do you think?

paper/manuscript.tex
```diff
@@ -208,7 +208,7 @@ \section*{Software availability}
 %% -------------------------------------------------------
 pykarambola is released under the GNU General Public License v3.0 or later (GPL-3.0-or-later) and is freely available at \url{https://github.com/Ishihara-SynthMorph/pykarambola}.
 The package can be installed via the Python Package Index (PyPI) with \texttt{pip install pykarambola}.
-Version 0.4.0, the version described in this paper, is permanently archived on Zenodo at \url{https://doi.org/10.5281/zenodo.20127022}.
+Version 0.3.0, the version described in this paper, is permanently archived on Zenodo at \url{https://doi.org/10.5281/zenodo.20127022}.
```
keisuke-ishihara: https://doi.org/10.5281/zenodo.20127022
this DOI is pointing to version 0.4.0

I think we should release 0.5.0 with with the w104 API change and then fix this line to v0.5.0 with the new DOI that will be minted.

keisuke-ishihara:
for the results plot, i like option A better because option B does not convey the improvement of pykarambola[accel] over the baseline. Also, always use number of vertices as the x-axis for mesh size (number of subdivision is algorithm dependent)

i find the time ratio a bit clunky to understand. how about we invert the y-axis and call it speed up?

image

keisuke-ishihara:
## Update: w104 API change isolated to separate PR

As discussed above, the `w104_eigvals` API change (closes #130) has been extracted into its own PR and release cycle to keep this benchmark PR clean.

**What was done:**
- PR #136 created with only the `w104_eigvals` addition to `api.py` + version bump `0.4.0 → 0.5.0` — merged to main
- GitHub release `v0.5.0` created and published to PyPI
- Zenodo archive: https://doi.org/10.5281/zenodo.20418801
- This PR updated: removed `Closes #130`, manuscript now references pykarambola v0.5.0 and the new Zenodo DOI

**Note for reviewer:** the benchmark was originally run against v0.3.0. Since `w104_eigvals` is now part of the comparison scope (and was computed from the same underlying `w104` Voigt matrix), please verify whether the pykarambola side of `benchmark_adrenal3d.ipynb` needs a re-run against v0.5.0 before merging.

keisuke-ishihara:
Investigation of #135 (MSM accuracy review) revealed that pykarambola returns `wl = 0` exactly for odd l by applying the Wigner 3j parity selection rule analytically in the Racah-formula implementation, while C++ karambola's `gsl_sf_coupling_3j` produces spurious residuals of order 1e-7 on biological meshes. This was traced back to the sympy → Racah-formula refactor, which preserved the analytical parity check that sympy had applied symbolically. Commit 49f449c adds a paragraph to the Numerical accuracy section of the manuscript documenting these findings.

paper/manuscript.tex
```diff
@@ -120,9 +120,8 @@ \subsection*{Runtime benchmark}
 Second, the AdrenalMNIST3D a standardized biomedical dataset~\cite{Yang2023MedMNIST} with 1{,}584 adrenal glands at two voxel resolutions (28$^3$ and 64$^3$) was used to measure performance on realistic biological meshes.
 All measurements were performed on a single core of an Apple~M4 processor (macOS~15.7.3, Python~3.13, pykarambola~v0.5.0).
 
-On icospheres, the Cython-accelerated build reduced compute time by ~1.1–1.2$\times$ relative to the pure-Python build (Figure~\ref{fig:runtime}).
-The absolute time saved by Cython grows with mesh size --- from $\sim$0.01~s at subdivision~5 to $\sim$26~s at subdivision~10.
-For real-world batch analyses — where hundreds or thousands of meshes are processed sequentially — the growing absolute time saving at larger mesh sizes translates directly into meaningful reductions in total pipeline runtime.
+On icospheres, the Cython-accelerated build reduced compute time by ~1.2–1.3$\times$ relative to the pure-Python build (Figure~\ref{fig:runtime}).
```
keisuke-ishihara: I think it's a bit weird that you refer to this speed up plot and start talking about the Cython acceleration first. The y-axis is a speed up of pykarambola relative to C++ Karambola.

It does not directly show the 1.2-1.3 times that you mention. I would suggest describing the blue line and orange lines first. Then make a small remark about Cython vs. pure python.

keisuke-ishihara:
image

removing these text will clarify the figure

keisuke-ishihara (changes requested):

paper/manuscript.tex
```diff
@@ -142,13 +142,10 @@ \subsection*{Numerical accuracy}
...
-The C++ implementation uses \texttt{gsl\_sf\_coupling\_3j}, which evaluates the coefficient numerically without an explicit parity check, producing spurious floating-point residuals of order $10^{-7}$--$10^{-8}$ on the AdrenalMNIST3D meshes (standard deviations across 1{,}584 shapes: $w_3$: $1.0 \times 10^{-7}$; $w_5$: $6.7 \times 10^{-8}$; $w_7$: $8.0 \times 10^{-8}$).
-The residuals arise because complex biological shapes have non-negligible odd-$l$ spherical harmonic coefficients $d_{lm}$ (of order $10^{-2}$--$10^{-1}$, consistent with $q_3 \approx 0.08$), so numerical errors in the 3j coefficients accumulate across the triple sum over $(m_a, m_b, m_c)$ rather than cancelling as they do for near-spherical geometries.
-This improvement was an indirect consequence of removing the \texttt{sympy} dependency: both the original \texttt{sympy} implementation and the Racah-formula replacement apply the parity rule analytically, whereas the C++ GSL backend does not.
+The spherical Minkowski summary quantities (\texttt{msm\_ql} and \texttt{msm\_wl}, $l = 0, \ldots, 8$) show a distinct accuracy pattern.
```
keisuke-ishihara: remember to reference the figure that shows numerical agreement

yajushikhurana: done!

yajushikhurana:
@keisuke-ishihara the PR is ready for review. 

1. the text was updated with rerun results with v0.5.0. The runtime can vary slightly with new runs (eg 11.1 to 11.3 seconds for the adrenal3d run). Either we should omit stating explicit time in manuscript or update manscript with new runtimes with any version updates prior to release. 
2. I chose to stick with the absolute difference in the text but I did add relative error plots. 
3. Careful review might be required for the benchmark_adrenal3d notebook. half of it was modified before my Claude code ran out today haha.

keisuke-ishihara:
@yajushikhurana 

## Figure "Runtime benchmark on synthetic icospheres"
- I think this needs to take up two columns (i.e. the full page width) to be legible.
- for panel a, you can drop (parse+compute) from the plot to save space. As long as this information is contained in the Figure legend, it'd be fine.

yajushikhurana:
> @yajushikhurana
> 
> ## Figure "Runtime benchmark on synthetic icospheres"
> * I think this needs to take up two columns (i.e. the full page width) to be legible.
> * for panel a, you can drop (parse+compute) from the plot to save space. As long as this information is contained in the Figure legend, it'd be fine.

Updated!

keisuke-ishihara (approved):

PR #152, opened 2026-07-03
docs: add MSM section to minkowski_tensors.md (closes #145)

yajushikhurana:
## Summary

- Adds a dedicated **Minkowski Structure Metrics** section to `docs/minkowski_tensors.md`, closing #145
- Covers `msm_ql` ($q_\ell$) and `msm_wl` ($w_\ell$) for $\ell = 0, \ldots, 8$
- Documents shared intermediate $D_{\ell m}$ accumulation (area-weighted spherical harmonics)
- Analytical and computational formulas for both metrics
- Parity selection rule: $w_\ell = 0$ exactly for odd $\ell$ (pykarambola Racah vs C++ karambola spurious residuals)
- Array shape note: `(12,)`, entries 0–8 meaningful
- Connection to `w104` and when independent face-normal evaluation is needed
- Adds **Mickel et al. (2013)** citation (J. Chem. Phys. 138, 044501) — the canonical MSM paper introducing Minkowski-tensor-based improvements over Steinhardt bond-order parameters
- Fixes GitHub LaTeX rendering: all heavy formulas moved to `$$` display blocks; `\begin{smallmatrix}` and `\substack` removed from inline/table-cell contexts; `\dfrac`/`\displaystyle` removed from table cells

## References added

- Steinhardt, Nelson & Ronchetti (1983) — original bond-order parameters
- Mickel, Kapfer, Schröder-Turk & Mecke (2013) — MSM introduction *(new)*
- Schröder-Turk et al. (2013) — Minkowski tensor review

### DIFF
--- a/docs/minkowski_tensors.md
+++ b/docs/minkowski_tensors.md
@@ -283,8 +283,8 @@ $$Q_{\ell m} = \frac{D_{\ell m}}{A\sqrt{\dfrac{4\pi}{2\ell+1}}}$$
 
 | | |
 |---|---|
-| **pykarambola key** | `msm_ql` (array shape `(12,)`, index $= \ell$, $\ell = 0,\ldots,8$) |
-| **Interpretation** | Strength of $\ell$-fold orientational order of surface normals. $q_0 = 1$ always; $q_\ell \approx 0$ for a smooth sphere; icosahedral symmetry yields a distinctive peak at $\ell = 6$. MSM $q_\ell$ and $w_\ell$ outperform the plain Steinhardt bond-order parameters for characterising disordered particulate matter (Mickel et al. 2013). |
+| **pykarambola key** | `msm_ql` (array shape `(12,)`, index $= \ell$, $\ell = 0,\ldots,8$; indices 9–11 are always zero padding) |
+| **Interpretation** | Strength of $\ell$-fold orientational order of surface normals. $q_0 = 1$ always; $q_\ell \approx 0$ for a smooth sphere; icosahedral symmetry yields a distinctive peak at $\ell = 6$. The series stops at $\ell = 8$ (`MAX_L = 8` in `spherical.py`), matching the default of the karambola C++ reference implementation. MSM $q_\ell$ and $w_\ell$ outperform the plain Steinhardt bond-order parameters for characterising disordered particulate matter (Mickel et al. 2013). |
 
 Analytical definition:
 
@@ -313,7 +313,7 @@ where the parenthesised quantity is the Wigner 3j symbol evaluated via the Racah
 
 | | |
 |---|---|
-| **pykarambola key** | `msm_wl` (array shape `(12,)`, index $= \ell$, $\ell = 0,\ldots,8$) |
+| **pykarambola key** | `msm_wl` (array shape `(12,)`, index $= \ell$, $\ell = 0,\ldots,8$; indices 9–11 are always zero padding) |
 | **Interpretation** | Phase structure of $\ell$-fold orientational order. $w_\ell = 0$ exactly for all odd $\ell$ (parity rule below). Distinguishes crystal symmetries that share the same $q_\ell$ value. |
 
 Analytical definition:
@@ -321,6 +321,12 @@ Analytical definition:
 $$
 w_\ell = \sqrt{\frac{4\pi}{2\ell+1}}\cdot\text{sgn}(\tilde{w}_\ell)\cdot|\tilde{w}_\ell|^{1/3}
 $$
+
+> **Convention note.** pykarambola's $w_\ell$ is the *unnormalized* cube-root form above and is
+> **not** divided by $q_\ell^3$. The Steinhardt (1983) and Mickel et al. (2013) convention is
+> $w_\ell^{\text{std}} = \tilde{w}_\ell / \bigl(\sum_m |Q_{\ell m}|^2\bigr)^{3/2}$.
+> These two quantities differ by a factor of $q_\ell^3$; numerical values will not match the
+> FCC/BCC tables in either paper when using pykarambola's output directly.
 
 Computational form: let
 
@@ -336,6 +342,9 @@ $$
 \mathtt{msm\_wl}[\ell] = \text{sgn}(\text{Re}(v))\cdot|v|^{1/3}/A
 $$
 
+($v$ is real for any closed mesh — imaginary parts cancel by the conjugation symmetry
+$D_{\ell,-m} = (-1)^m\,\overline{D_{\ell m}}$ — so $\operatorname{sgn}(\operatorname{Re}(v)) = \operatorname{sgn}(v)$.)
+
 **Parity selection rule.** For odd $\ell$, the Wigner 3j symbol above is antisymmetric
 under exchange of any two columns (each exchange contributes $(-1)^{3\ell}=-1$),
 while $Q_{\ell m_1}Q_{\ell m_2}Q_{\ell m_3}$ is symmetric.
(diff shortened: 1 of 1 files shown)


yajushikhurana:
Requested for review @keisuke-ishihara

keisuke-ishihara (commented):
The w202 coefficient change (24→12) is correct — the code traverses each edge via two adjacent triangles at \`/24\` each, netting \`/12\` per edge in the single-sum formula; three box-geometry tests confirm this. The new MSM section is generally clear and well-structured. Eight issues below, roughly severity-ordered.

---

### 1. \`msm_wl\` normalization deviates from the cited Steinhardt/Mickel convention without noting it — line 322

The doc defines:

$$w_\ell = \sqrt{\frac{4\pi}{2\ell+1}} \cdot \operatorname{sgn}(\tilde{w}_\ell) \cdot |\tilde{w}_\ell|^{1/3}$$

The Steinhardt (1983) and Mickel et al. (2013) convention is:

$$w_\ell = \frac{\tilde{w}_\ell}{\left(\sum_m |Q_{\ell m}|^2\right)^{3/2}}$$

These are different quantities — pykarambola's version is not normalized by $q_\ell^3$. A user comparing \`msm_wl\` output against the FCC/BCC tabulated values in either paper will find numerical disagreement and incorrectly conclude there is a bug. The doc cites both papers as defining this metric without flagging the deviation.

**Suggestion:** Add a note that pykarambola's $w_\ell$ is the unnormalized cube-root form and differs from the Steinhardt/Mickel definition by a factor of $q_\ell^3$.

---

### 2. \`J % 2 != 0\` shortcut is described as a correct analytic selection rule — it isn't — line 346

> pykarambola enforces this analytically via the \`J % 2 != 0\` branch of the Racah formula.

\`J % 2 != 0\` is not a Wigner 3j selection rule. For example, $\begin{pmatrix}1&1&1\\1&-1&0\end{pmatrix} \approx 0.408 \neq 0$ even though $J = 3$ is odd. The shortcut zeroes every individual symbol with odd $J$ as a bulk early-return, which happens to give the correct sum for \`msm_wl\` (because the full sum over $(m_1,m_2,m_3)$ cancels by antisymmetry), but returns wrong values for any other caller of \`_wigner3j\` with odd $j_1+j_2+j_3$. This is also the root cause of the bispectrum incompleteness in issue #137.

Describing the shortcut as an analytic enforcement of a selection rule presents a known bug as a feature.

**Suggestion:** Replace the sentence with something like: "For odd $\ell$, every term in the sum cancels pairwise by antisymmetry, so the sum is never evaluated. Note: the internal \`_wigner3j\` helper uses a J-parity early return that is only correct within the \`msm_wl\` context — see #137 for the known limitation."

---

### 3. \`W_1^{0,\ell}\` connection formula is missing the 1/3 prefactor — line 356

$$W_1^{0,\ell} = \sum_{f \in F} \underbrace{\hat{n}_f \otimes \cdots \otimes \hat{n}_f}_{\ell \text{ times}} A_f$$

Every other $W_1^{0,s}$ definition in this file (w102 line 214, w103 line 242, w104 line 251) includes the $\frac{1}{3}$ Steiner prefactor. A reader using this formula to cross-check against pykarambola output will find a factor-of-3 discrepancy.

**Suggestion:** $W_1^{0,\ell} = \dfrac{1}{3}\displaystyle\sum_{f \in F} \hat{n}_f^{\otimes\ell}\, A_f$

---

### 4. Mickel et al. 2013 is listed twice in References — line 396

Lines 389–393 and 396–399 are duplicate entries for the same paper (same authors, same DOI: \`10.1063/1.4774084\`). The first reads "for the study of disordered particulate matter"; the second reads "for the analysis of disordered particulate matter". One must be deleted.

---

### 5. \`sgn(Re(v))\` in the computational form is unexplained — line 336

The analytical definition uses $\operatorname{sgn}(\tilde{w}_\ell)$, which is real. The computational form uses $\operatorname{sgn}(\operatorname{Re}(v))$ where $v$ is a sum of products of complex $D_{\ell m}$ values. The doc never establishes when $\operatorname{Re}(v)$ carries the correct sign.

**Suggestion:** Add a parenthetical: "$v$ is real for any closed mesh (imaginary parts cancel by the conjugation symmetry $D_{\ell,-m} = (-1)^m \overline{D_{\ell m}}$), so $\operatorname{sgn}(\operatorname{Re}(v)) = \operatorname{sgn}(v)$."

---

### 6. Array shape \`(12,)\` with indices 9–11 silently zero, and \`lmax = 8\` rationale not explained — lines 286, 316

Both \`msm_ql\` and \`msm_wl\` are documented as "array shape \`(12,)\`, index $= \ell$, $\ell = 0,\ldots,8$". Two things are missing:

- The three unused tail slots are never explained. A user iterating over all 12 entries may interpret indices 9–11 as $\ell = 9, 10, 11$ results with zero orientational order. The \`12\` appears to be an implementation artifact (no named constant, no comment in the code); it is not \`lmax + 1\` (that would be 9) or any other obvious quantity.
- The choice of \`lmax = 8\` (\`MAX_L = 8\` in \`spherical.py\`) is unexplained. The reason is that $\ell = 8$ is the lowest order at which FCC and HCP packings can be distinguished by their $q_\ell$ spectrum — context that originates from the crystal lattice characterisation literature (Steinhardt 1983; Mickel et al. 2013). Issue #145 also explicitly requested this explanation.

**Suggestion:** Add "(indices 9–11 are always zero padding)" to both table rows, and add a sentence explaining that the series stops at $\ell = 8$ because that is the lowest order needed to distinguish FCC from HCP.

---

### 7. Relationship to \`w104\` is left ambiguous — lines 350–360

The doc says MSM and Cartesian moment tensors $W_1^{0,\ell}$ "are different representations of the same Minkowski tensors and can therefore be related to each other." This is true mathematically, but it leaves open the question of whether pykarambola computes MSM *from* \`w104\`. It does not: \`SphericalMinkowskis\` in \`spherical.py\` accumulates $D_{\ell m}$ directly from face normals, independently of the \`w104\` code path. Issue #145 originally assumed MSM were "derived from \`w104\` via contraction with Wigner 3j symbols"; that assumption was silently dropped without explanation.

**Suggestion:** Add a sentence making the independence explicit, e.g.: "In pykarambola, \`msm_ql\` and \`msm_wl\` are computed via a separate spherical harmonic accumulation (\`spherical.py\`) and share no code with the \`w104\` computation path."

(The last sentence of the section already says this in slightly weaker form — it just needs to be made more prominent, or the "related to each other" claim needs a caveat that the relationship is mathematical, not computational.)

yajushikhurana:
**Re: issue #6 — `lmax = 8` rationale**

The suggestion to explain `MAX_L = 8` as "the lowest order needed to distinguish FCC from HCP" appears to be incorrect. Mickel et al. (2013) Table I shows FCC and HCP already differ at ℓ = 3 (q₃ = 0 for FCC by inversion symmetry, q₃ = 0.076 for HCP) and plainly at ℓ = 4 (q₄ = 0.190 vs 0.097). We've updated the doc to say the cutoff matches the karambola C++ default instead.

Could you provide a citation for the ℓ = 8 / FCC–HCP claim? If there's a specific paper or quantity (e.g. w_ℓ values rather than q_ℓ) that makes ℓ = 8 special for that discrimination, we're happy to reinstate it with the correct reference.

keisuke-ishihara:
Confirmed. Mickel et al. (2013) Table I shows FCC and HCP already differ at ℓ = 3 (q₃ = 0 vs 0.076) and clearly at ℓ = 4 (q₄ = 0.190 vs 0.097), so ℓ = 8 is not the minimum for that discrimination. There is no citation to provide — the original suggestion was incorrect. The current text ("matching the default of the karambola C++ reference implementation") is accurate and sufficient. No change needed.

keisuke-ishihara:
I think the doc is complete. Merging!

PR #153, opened 2026-07-06
feat: make MSM max_l configurable (default 8, matching C++ karambola)

keisuke-ishihara:
## Summary

- Adds `msm_max_l` parameter to `minkowski_tensors` and `minkowski_tensors_from_label_image` (default 8, preserving existing behavior)
- Adds `--msm-max-l` CLI flag
- `msm_ql`/`msm_wl` output arrays always have at least 12 elements; grow beyond 12 only when `msm_max_l >= 12`
- Output file writer (`write_sphmink_file`) now generates column headers dynamically instead of hardcoding 9 columns
- Removes dead `MAX_L = 8` module-level constant from `spherical.py`

## Follow-up

After this lands, `docs/msm-section-minkowski-tensors-md` (#145) needs to be updated:
- The array shape description currently says `(12,)` — it should reflect the dynamic shape `(max(12, msm_max_l + 1),)` and note that indices 0–`msm_max_l` are populated
- The ℓ = 0,...,8 upper bound in the key table should become ℓ = 0,...,`msm_max_l`

## Test plan

- [ ] Existing tests pass (`msm_ql`/`msm_wl` keys present and non-empty)
- [ ] Verify default output shape is still `(12,)` with `msm_max_l=8`
- [ ] Verify `msm_max_l=11` produces shape `(12,)` (capped at 12)
- [ ] Verify `msm_max_l=14` produces shape `(15,)` (grows beyond 12)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ Version numbers follow [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
-### Changed
-- `MAX_L` raised from 8 to 11 in `spherical.py`, enabling MSM computation up to l=11. `SphMinkData` already allocated 12 slots (indices 0–11); only the loop bound was increased.
+### Added
+- `msm_max_l` parameter on `minkowski_tensors` and `minkowski_tensors_from_label_image` (default 8, matching C++ karambola) to control the maximum spherical harmonic order for MSM. Output arrays `msm_ql`/`msm_wl` always have at least 12 elements (preserving existing shape); they grow beyond 12 only when `msm_max_l >= 12`. The `--msm-max-l` flag exposes this option on the CLI.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Version numbers follow [Semantic Versioning](https://semver.org/).
 ## [Unreleased]
 
 ### Added
-- `msm_max_l` parameter on `minkowski_tensors` and `minkowski_tensors_from_label_image` (default 8, matching C++ karambola) to control the maximum spherical harmonic order for MSM. Output arrays `msm_ql`/`msm_wl` always have at least 12 elements (preserving existing shape); they grow beyond 12 only when `msm_max_l >= 12`. The `--msm-max-l` flag exposes this option on the CLI.
+- `msm_max_l` parameter on `minkowski_tensors` and `minkowski_tensors_from_label_image` (default 8, matching C++ karambola) to control the maximum spherical harmonic order for MSM. Output arrays `msm_ql`/`msm_wl` have shape `(msm_max_l + 1,)`. The `--msm-max-l` flag exposes this option on the CLI.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Version numbers follow [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+- `MAX_L` raised from 8 to 11 in `spherical.py`, enabling MSM computation up to l=11. `SphMinkData` already allocated 12 slots (indices 0–11); only the loop bound was increased.
+
 ---
 
 ## [0.5.1] - 2026-06-14
--- a/docs/minkowski_tensors.md
+++ b/docs/minkowski_tensors.md
@@ -283,8 +283,8 @@ $$Q_{\ell m} = \frac{D_{\ell m}}{A\sqrt{\dfrac{4\pi}{2\ell+1}}}$$
 
 | | |
 |---|---|
-| **pykarambola key** | `msm_ql` (array shape `(12,)`, index $= \ell$, $\ell = 0,\ldots,8$; indices 9–11 are always zero padding) |
-| **Interpretation** | Strength of $\ell$-fold orientational order of surface normals. $q_0 = 1$ always; $q_\ell \approx 0$ for a smooth sphere; icosahedral symmetry yields a distinctive peak at $\ell = 6$. The series stops at $\ell = 8$ (`MAX_L = 8` in `spherical.py`), matching the default of the karambola C++ reference implementation. MSM $q_\ell$ outperforms the plain Steinhardt bond-order parameters for characterising disordered particulate matter (Mickel et al. 2013). |
+| **pykarambola key** | `msm_ql` (array shape `(max(12, msm_max_l + 1),)`, index $= \ell$, $\ell = 0,\ldots,\texttt{msm\_max\_l}$; indices beyond `msm_max_l` are always zero padding) |
+| **Interpretation** | Strength of $\ell$-fold orientational order of surface normals. $q_0 = 1$ always; $q_\ell \approx 0$ for a smooth sphere; icosahedral symmetry yields a distinctive peak at $\ell = 6$. The default `msm_max_l=8` matches the karambola C++ reference implementation; pass a larger value to extend the series. MSM $q_\ell$ outperforms the plain Steinhardt bond-order parameters for characterising disordered particulate matter (Mickel et al. 2013). |
 
 Analytical definition:
 
@@ -313,7 +313,7 @@ where the parenthesised quantity is the Wigner 3j symbol evaluated via the Racah
 
 | | |
 |---|---|
-| **pykarambola key** | `msm_wl` (array shape `(12,)`, index $= \ell$, $\ell = 0,\ldots,8$; indices 9–11 are always zero padding) |
+| **pykarambola key** | `msm_wl` (array shape `(max(12, msm_max_l + 1),)`, index $= \ell$, $\ell = 0,\ldots,\texttt{msm\_max\_l}$; indices beyond `msm_max_l` are always zero padding) |
 | **Interpretation** | Phase structure of $\ell$-fold orientational order. $w_\ell = 0$ exactly for all odd $\ell$ (parity rule below). Distinguishes crystal symmetries that share the same $q_\ell$ value. |
 
 Analytical definition:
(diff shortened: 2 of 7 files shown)


yajushikhurana:
## Review

All 217 existing tests pass on this branch. The three test-plan scenarios also work correctly when verified manually. A few issues below.

---

### 1. No tests added — test plan checkboxes are all unchecked

The three scenarios should become actual test assertions:
- `msm_max_l=8` (default) → shape `(12,)`
- `msm_max_l=11` → shape `(12,)` (capped at minimum)
- `msm_max_l=14` → shape `(15,)` (grows beyond minimum)

---

### 2. `output.py`: `n_l` should be `max_l + 1`, not `len(ql)`

```python
# current
n_l = len(next(iter(sphmink.values())).result.ql) if sphmink else 12
```

With `max_l=8` (default), `len(ql) = 12`, so the file writer loops `range(12)` and emits **12 q/w column pairs** — including 3 zero-padded slots — instead of the previous **9**. This silently changes the output file format for every existing user on the default invocation.

The right number of columns to write is `max_l + 1` (only the meaningful orders). The 12-slot in-memory array is an internal compatibility detail for matching the C++ karambola array shape; there is no reason for it to appear in the file format. `SphMinkData` would need to store `max_l` as an attribute so `write_sphmink_file` can read it back.

Same issue on the data loop:
```python
for i in range(len(result.result.ql)):   # writes 12 pairs for default max_l=8
```
should be `range(result.result.max_l + 1)`.

Once we decide the length we also need to update documentation for PR #152 .

---

### 3. `msm_max_l < 0` silently produces incorrect results

```python
minkowski_tensors(..., msm_max_l=-1)  # ql = [0, 0, 0, ...]  (q0 should always be 1)
```

With `max_l=-1`, the allocation loop `range(0)` never runs so `q0 = 0` rather than the correct `1.0`. A `ValueError` guard on `msm_max_l < 0` in `api.py` and `calculate_sphmink` would prevent this.

keisuke-ishihara (commented):
Thanks for the thorough review — all three issues addressed in fe3c9fa.

**1. Tests added** — four new tests in `TestComputeOptions`:
- `msm_max_l=8` (default) → shape `(12,)`
- `msm_max_l=11` → shape `(12,)` (floored)
- `msm_max_l=14` → shape `(15,)` (grows)
- `msm_max_l=-1` → `ValueError`

**2. `output.py` column count fixed** — `SphMinkData` now stores `self.max_l` as an attribute. `write_sphmink_file` uses `result.result.max_l + 1` for both the header and data loops, so the default invocation writes 9 q/w column pairs again (matching the original C++ format). The 12-slot internal array is an implementation detail that no longer leaks into the file.

**3. Negative `max_l` guard added** — `ValueError` raised in both `calculate_sphmink` and `minkowski_tensors` for `max_l < 0` / `msm_max_l < 0`.

Also updated `docs/minkowski_tensors.md` to reflect the dynamic array shape `(max(12, msm_max_l + 1),)` and configurable upper bound for both `msm_ql` and `msm_wl` (as noted in your comment about PR #152).

yajushikhurana (commented):
Four issues found on the current branch state:

---

**1. `w202` formula corrupted by merge (bug)**

Commit `270c5a7` (merge of `main` into this branch) accidentally changed the `w202` formula in `docs/minkowski_tensors.md` from `/24` to `/12`:

```
-w_{202,ij} = \sum_e \left[ \frac{\ell_e(\alpha_e + \sin\alpha_e)}{24} ... + \frac{\ell_e(\alpha_e - \sin\alpha_e)}{24} ... \right]
+w_{202,ij} = \sum_e \left[ \frac{\ell_e(\alpha_e + \sin\alpha_e)}{12} ... + \frac{\ell_e(\alpha_e - \sin\alpha_e)}{12} ... \right]
```

The implementation at `minkowski.py:513–514` still correctly uses `/24.0`. The docs need to be reverted to `/24`.

---

**2. MSM section intro hardcodes `ℓ = 0,...,8` (stale after this PR)**

```
Both are computed from area-weighted spherical harmonic moments of face normals
for ℓ = 0,...,8 and returned only with compute='all'.
```

Should now read `ℓ = 0,...,msm_max_l` (or similar) since the upper bound is configurable.

---

**3. `SphMinkData` should use `max_l + 1`, not `max(12, max_l + 1)`**

The 12-slot floor was inherited from a last-minute C++ karambola implementation detail (they initially intended to release with `lmax=12` but shipped with `lmax=8`, leaving the array oversized). There is no mathematical or physical reason for it.

Keeping the floor now creates an inconsistency: the Python API returns shape `(12,)` for the default `msm_max_l=8`, but the output file correctly writes 9 column pairs. A user reading the file and the array would see different lengths.

Suggested change in `spherical.py`:
```python
# Before
size = max(12, max_l + 1)

# After
size = max_l + 1
```

The two tests that currently assert `shape == (12,)` for `max_l <= 11` would need updating to `(9,)` and `(12,)` respectively. Since this is an unreleased feature there is no backwards-compatibility cost, and the CHANGELOG note about the 12-element floor can be simplified.

---

**4. CHANGELOG wording is ambiguous (minor)**

> Output arrays `msm_ql`/`msm_wl` always have at least 12 elements (preserving existing shape); they grow beyond 12 only when `msm_max_l >= 12`.

The word "output" is ambiguous — "output arrays" reads as the Python return values, but the output *file* now writes `max_l+1` pairs (9 for default), not 12. If issue 3 is addressed, the whole clause about the 12-element floor goes away and the ambiguity resolves itself.

yajushikhurana:
Four issues found on the current branch state:

---

**1. `w202` formula corrupted by merge (bug)**

Commit `270c5a7` (merge of `main` into this branch) accidentally changed the `w202` formula in `docs/minkowski_tensors.md` from `/24` to `/12`:

```
-w_{202,ij} = \sum_e \left[ \frac{\ell_e(\alpha_e + \sin\alpha_e)}{24} ... + \frac{\ell_e(\alpha_e - \sin\alpha_e)}{24} ... \right]
+w_{202,ij} = \sum_e \left[ \frac{\ell_e(\alpha_e + \sin\alpha_e)}{12} ... + \frac{\ell_e(\alpha_e - \sin\alpha_e)}{12} ... \right]
```

The implementation at `minkowski.py:513–514` still correctly uses `/24.0`. The docs need to be reverted to `/24`.

---

**2. MSM section intro hardcodes `ℓ = 0,...,8` (stale after this PR)**

```
Both are computed from area-weighted spherical harmonic moments of face normals
for ℓ = 0,...,8 and returned only with compute='all'.
```

Should now read `ℓ = 0,...,msm_max_l` (or similar) since the upper bound is configurable.

---

**3. `SphMinkData` should use `max_l + 1`, not `max(12, max_l + 1)`**

The 12-slot floor was inherited from a last-minute C++ karambola implementation detail (they initially intended to release with `lmax=12` but shipped with `lmax=8`, leaving the array oversized). There is no mathematical or physical reason for it.

Keeping the floor creates an inconsistency: the Python API returns shape `(12,)` for the default `msm_max_l=8`, but the output file correctly writes 9 column pairs. A user reading the file and the array would see different lengths.

Suggested change in `spherical.py`:
```python
# Before
size = max(12, max_l + 1)

# After
size = max_l + 1
```

The two tests asserting `shape == (12,)` for `max_l <= 11` would need updating to `(9,)` and `(12,)` respectively. Since this is an unreleased feature there is no backwards-compatibility cost, and the CHANGELOG note about the 12-element floor can be simplified.

---

**4. CHANGELOG wording is ambiguous (minor)**

> Output arrays `msm_ql`/`msm_wl` always have at least 12 elements (preserving existing shape); they grow beyond 12 only when `msm_max_l >= 12`.

"Output arrays" is ambiguous — it reads as the Python return values, but the output *file* already writes `max_l+1` pairs (9 for default), not 12. If issue 3 is addressed, the clause about the 12-element floor disappears and the ambiguity resolves itself.

keisuke-ishihara:
**Re: issue 1 — w202 `/24` vs `/12`**

The `/12` in the docs is correct and was not changed by the merge.

The code (`minkowski.py:513–514`) accumulates `ℓ_e (α_e ± sin α_e) / 24` **per triangle per edge-slot**. Each edge is shared by exactly two triangles, so each edge is visited twice, and the net per-edge contribution is `2 × (1/24) = 1/12`. The displayed formula sums over edges (one term per edge), so the `/12` denominator is right.

You can verify with `git show 1a6a0fe:docs/minkowski_tensors.md | grep "w_{202"` — the formula already read `/12` in the original PR #152 commit, and `git diff 1a6a0fe..270c5a7 -- docs/minkowski_tensors.md` shows no change to that line. No correction needed here.

yajushikhurana (commented):
**Apology for previous review — w202 formula is correct**

My earlier comment flagging the `w202` formula as a merge artifact was wrong. The `/12` prefactor was an intentional fix introduced in PR #152 (commit `b5adfcb`: *"fix w202 formula prefactor: 1/12 (edge sum) not 1/24 (triangle loop)"*) and is already on `main`. Sorry for the noise.

---

**Remaining issue — stale docstring in `api.py` (blocker)**

The `msm_max_l` docstring at `api.py:285–289` still describes the old 12-element floor that was removed in the last commit (`adb5a5`):

> The output arrays `msm_ql` and `msm_wl` always have at least 12 elements (indices 0–11); they grow beyond 12 only when `msm_max_l >= 12`.

Should be updated to match the actual behavior:

> Output arrays `msm_ql` and `msm_wl` have shape `(msm_max_l + 1,)`.

yajushikhurana (commented):
**Apology for previous review — w202 formula is correct**

My earlier comment flagging the `w202` formula as a merge artifact was wrong. The `/12` prefactor was an intentional fix introduced in PR #152 (commit `b5adfcb`: *"fix w202 formula prefactor: 1/12 (edge sum) not 1/24 (triangle loop)"*) and is already on `main`. Sorry for the noise.

---

**Remaining issue — stale docstring in `api.py` (blocker)**

The `msm_max_l` docstring at `api.py:285–289` still describes the old 12-element floor that was removed in the last commit (`adb5a5`):

> The output arrays `msm_ql` and `msm_wl` always have at least 12 elements (indices 0–11); they grow beyond 12 only when `msm_max_l >= 12`.

Should be updated to match the actual behavior:

> Output arrays `msm_ql` and `msm_wl` have shape `(msm_max_l + 1,)`.

yajushikhurana (approved):
All feedback addressed. LGTM.

PR #157, opened 2026-08-01
perf: crop to bbox in minkowski_tensors_from_label_image (#156)

keisuke-ishihara:
Closes #156.

## Summary

- Pre-compute per-label bounding boxes via `skimage.measure.regionprops` in one O(image_size) pass, then crop to each object's bbox (+ 1 voxel padding) before creating the float64 mask and calling `marching_cubes`.
- Eliminates the O(N_labels × image_size) bottleneck that caused `minkowski_tensors_from_label_image` to scan and allocate a full-image float64 array for every label.
- Applied to all three code paths with the bottleneck: the main per-label loop, the `return_count` loop, and the global `centroid_mesh` pre-pass.

## Performance

On a representative 77×576×576 label image (4334 labels after border removal, 2× downsampled from 154×1152×1152):

| | Before | After |
|---|---|---|
| Time per label | ~136 ms | ~10 ms |
| Extrapolated total | ~10 min | ~0.7 min |
| **Speedup** | | **~13×** |

The `regionprops` bbox precompute costs 0.07 s for all 4334 labels.

## API

No changes to the public API — same function signature, same return type, numerically identical outputs. All 203 tests pass.

## Test plan

- [x] `pytest tests/test_api.py::TestLabelImage` — 21 tests pass
- [x] `pytest` (full suite) — 203 passed, 18 skipped, 0 failures

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ Version numbers follow [Semantic Versioning](https://semver.org/).
 ### Added
 - `msm_max_l` parameter on `minkowski_tensors` and `minkowski_tensors_from_label_image` (default 8, matching C++ karambola) to control the maximum spherical harmonic order for MSM. Output arrays `msm_ql`/`msm_wl` have shape `(msm_max_l + 1,)`. The `--msm-max-l` flag exposes this option on the CLI.
 
+### Performance
+- `minkowski_tensors_from_label_image` now pre-computes per-label bounding boxes via `skimage.measure.regionprops` (one O(image_size) pass) and crops the image to each object's bounding box before calling `marching_cubes`. This eliminates the previous O(N_labels × image_size) cost and yields a ~13× wall-time speedup on a representative 77×576×576 image with 4334 labels (from ~10 min to ~0.7 min). (#156)
+
 ---
 
 ## [0.5.1] - 2026-06-14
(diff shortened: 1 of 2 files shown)


yajushikhurana (commented):
## Review

Overall this is a clean, well-scoped fix. The regionprops precompute + bbox crop approach is correct and the coordinate restoration (`v + crop_origin - _pad_offset`) is applied consistently across all three code paths. 203 tests passing with no API changes — looks good to merge.

**One optional suggestion** (`api.py`, `_crop_label` else-branch, ~line 819):

The fallback when `lab_int` is not in `_label_bboxes` silently returns the full image as the crop. This path shouldn't be reachable in normal flow (since `unique_labels` is derived from the image itself), but if it ever were hit — e.g. due to a future refactor or an edge case with the `labels=` argument — it would silently regress to the O(image_size) behaviour with no indication of what went wrong.

Suggest adding a `warnings.warn` there so the failure mode is visible:

```python
else:
    warnings.warn(
        f"Label {lab_int} not found in regionprops bounding-box cache; "
        "falling back to full-image crop (performance may degrade). "
        "This is unexpected — please file a bug report.",
        RuntimeWarning,
        stacklevel=3,
    )
    z0, y0, x0 = 0, 0, 0
    z1, y1, x1 = _shape
```

Not a blocker — just a small defensive improvement.

PR #159, opened 2026-08-01
perf: vectorize _get_open_and_nonmanifold fan traversal (#158)

keisuke-ishihara:
## Summary

- Pre-builds an `entry_succ` array (shape `(3F,)`) in one O(F) NumPy pass: for every vertex–triangle CSR entry, the fan-next triangle obtained by crossing the edge of that triangle at that vertex.
- The per-vertex loop then replaces ~5 Python method calls + a generator expression per step with a dict lookup into the pre-built array.
- Falls back gracefully to the non-CSR (`_vertex_triangles` list) path when `_accel` is not compiled.

## Benchmark (sphere meshes, 200 reps)

| r | V | F | Old (ms) | New (ms) | Speedup |
|---|---|---|----------|----------|---------|
| 4 | 294 | 584 | 1.82 | 0.56 | 3.2× |
| 7 | 894 | 1784 | 5.80 | 1.91 | 3.0× |
| 10 | 1902 | 3800 | 13.0 | 4.3 | 3.0× |
| 15 | 4254 | 8504 | 27.1 | 8.8 | 3.1× |

On the representative 77×576×576 label image (4334 nuclei), `_get_open_and_nonmanifold` was previously ~2.7 ms/label (~45% of total pykarambola time). The ~3× reduction brings it to ~0.9 ms/label.

## Test plan

- [x] All 213 existing tests pass (including `TestNonManifoldMesh::test_warns_non_manifold` with the bowtie mesh and all `TestOpenSurface` cases)
- [x] Both CSR (Cython `_accel`) and pure-Python fallback paths exercised

Closes #158

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Version numbers follow [Semantic Versioning](https://semver.org/).
 
 ### Performance
 - `minkowski_tensors_from_label_image` now pre-computes per-label bounding boxes via `skimage.measure.regionprops` (one O(image_size) pass) and crops the image to each object's bounding box before calling `marching_cubes`. This eliminates the previous O(N_labels × image_size) cost and yields a ~13× wall-time speedup on a representative 77×576×576 image with 4334 labels (from ~10 min to ~0.7 min). (#156)
+- `_get_open_and_nonmanifold` (called once per `minkowski_tensors` invocation) now uses a vectorized NumPy fan-traversal: the per-vertex loop replaces ~5 Python method calls per edge step with a pre-built `entry_succ` array and a local dict lookup, eliminating generator overhead. This reduces the non-manifold check from ~2.7 ms/label to sub-millisecond on typical nuclei. (#158)
 
 ---
 
--- a/pykarambola/api.py
+++ b/pykarambola/api.py
@@ -174,37 +174,75 @@ def _get_open_and_nonmanifold(surface):
         else set()
     )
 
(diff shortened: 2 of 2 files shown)


yajushikhurana (commented):
## Review

Clean, well-structured speedup. The `entry_succ` precompute is correct (`nb[vt_idx, entry_slot]` is exactly what the old per-step method calls were computing), the fallback path for non-CSR builds is handled, and the dead `if is_non_manifold: break` is correctly removed. 213 tests passing.

**One optional suggestion** — the removed integrity guard:

The old code had:
```python
assert vid != -1, f"Vertex {i} not found in any slot of triangle {cur_t}"
```

The new code uses `np.argmax(..., axis=1)` to find the slot, but `np.argmax` silently returns `0` on no match. If CSR data were ever corrupted, the old code would crash with a clear error; the new code would silently produce a wrong result.

Since the CSR structure is internally consistent by construction this is unlikely to matter in practice — but the safety net is gone. Could replace it with a vectorized check right after computing `entry_slot`:

```python
entry_slot = np.argmax(
    faces_arr[vt_idx] == entry_vert[:, None], axis=1
)
slot_found = np.any(faces_arr[vt_idx] == entry_vert[:, None], axis=1)
if not np.all(slot_found):
    raise AssertionError("CSR integrity error: vertex not found in its own triangle.")
```

Not a blocker — just restores the original defensive check in vectorized form.

